PR #30, opened 2013-11-21
primitive outline for diary generation \w\i

btwe:
This patch improves the formatting of the diary overview.  Each diary
file is scanned and all occurences of 2nd level headings are printed
below the link entry to that diary file.  Thus, one can see what has
happenend on that day.

TODO: configure which heading levels should be accounted.


hq6:
Since the creator of this PR seems no longer interested in polishing it, and it does not appear to be complete enough to merge, I am closing this PR in the interest of repo sanitation.

PR #47, opened 2014-05-04
further large table optimizations

lyokha:
_(Please disregard changes in README.md: i do not know how did they get here and how to get rid of them.)_
1. `s:get_aligned_rows()`: getting 2 last rows is enough for having been formatted tables: this makes adding new rows in large tables really fast
2. `vimwiki#tbl#get_cells()`: using faster `strpart()` instead concatenating every new character into variables `cell` and `quote`
3. checking by `getline()` whether the line was changed before `setline()` does matter on slower computers


hq6:
I have recently been added as a new maintainer. 
Since it's been five years, I cannot ascertain whether this PR is still relevant / useful. @lyokha , if you believe this PR is still relevant, please resolve the conflicts and I will test it and try to get it merged in. 

Otherwise, I will close this PR in a few days under the assumption that it is outdated.

lyokha:
@hq6, I resolved conflicts and removed all changes that mistakenly appeared in *README* (but this was surprisingly painful, and I made 3 commits for that). Also I added docs about the new fast-table-format commands *gq1* and *gw1*.

lyokha:
I tested the changes, everything looks Ok.

hq6:
@lyokha , For posterity, can you provide some examples of tables you've tested this with?

lyokha:
It works with any Vimwiki table. Say, you open a new file

```ShellSession
$ vim tmp.wiki
```

Then in vim type command

```
:VimwikiTable
```

You'll see

```

|   |   |   |   |   |
|---|---|---|---|---|
|   |   |   |   |   |
```

Go to, say, in the first column in the header, press `i` for Insert mode, and print something, say, *Col1*, then move cursor to the 1st data row, 2nd column and type *data1*: the table is not aligned at this moment, then press `Enter`: another row will be inserted and the table will align, while the cursor will move at the 2nd row, 2nd column. Now the table looks like

```

| Col1 |       |   |   |   |
|------|-------|---|---|---|
|      | data1 |   |   |   |
|      |       |   |   |   |
```

So, basically formatting works. Now press `Esc` to enter the Normal mode, say `yy` and then `20p`: you'll get aligned table with 22 rows (which are empty from the 2nd to the bottom). Now select *data1* (with `viwy` or some other command), go to, say, 2nd row, 3rd column and press `p` to paste it there. Now the table looks misaligned:

```
| Col1 |       |   |   |   |
|------|-------|---|---|---|
|      | data1 |   |   |   |
|      |       | data1  |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
```

Let's align it. Move cursor to, say, tenth column and press `gq1`: nothing happens. Now press `gqq`: the table gets aligned. That's ok. Now press `u` to revert to mis-alignment, move the cursor to the 2nd or the third column (i.e. on the misaligned line or a line below it), and press `gq1`: now the table gets aligned.

Command `gq1` shows how the new optimization really works: when in the Insert mode you are inserting a new row, only a few lines above the current line get tested for mis-alignment: this is really enough for normal editing flow. Before the optimization, all lines in a table were checked, and the check slowed down depending on height of the table.

Historically, I used Vimwiki tables for dictionaries like it was explained [here](https://github.com/lyokha/vim-xkbswitch#custom-keyboard-layout-switching-rules). So, when a table got large, delays due to the not optimized table formatting algorithm were really annoying. With this optimization, huge tables must align much faster and, which is even more important, now they do not depend on table heights.

lyokha:
Hmm, actually the optimization is not working (long tables get slower), probably @EinfachToll was right about regression bug. I'll try to sort it out.

hq6:
Thanks for testing this again and discovering the performance issues! 
Please update this PR when you've resolved the performance regression.

lyokha:
@hq6, I sorted out the issue with no performance gain. The issue was that an autocmd on `InsertLeave` was using `vimwiki#tbl#format(line("."))` without the additional argument (this autocmd is set when `g:vimwiki_table_auto_fmt` is on), and this was triggering when inserting a new row in a table in function `s:kbd_create_new_row()`, because its algorithm requires leaving Insert mode for calling commands in Normal mode.

hq6:
@lyokha , When I put my cursor on line 6 of the following table and use `gq1`, it result is that only lines 4 to 6 are aligned. Is that the desired behavior?

```
| Col1 |       |   |   |   |
|------|-------|---|---|---|
|      | data1 |   |   |   |
|      |       | data1  |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
```

Output after `gq1` while cursor is on line 6.

```
| Col1 |       |   |   |   |
|------|-------|---|---|---|
|      | data1 |   |   |   |
|      |       | data1  |   |   |
|      |       |        |   |   |
|      |       |        |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
|      |       |   |   |   |
```

lyokha:
Yes, this is expected behavior. The algorithm is as follows. `gq1` has a *viewport*: it only sees 2 lines above the current line and the current line itself (however, the current line gets preliminary *formatted*, see details below). If the first line in the viewport is the longest, then other two lines are aligned with the length of the first line. If in the viewport there is a line which is longer than the first of the three, then the algorithm falls back to the original not-optimized algorithm, i.e. all lines in the table get formatted.

This is what you see: when pressing `gq1` on lines 4 and 5, the whole table should get formatted with not-optimized algorithm, on line 6 the optimized algorithm triggers (notice also, that in this case the line 4 does not shrink as it would have shrink if the whole table would get formatted: it contains a trailing whitespace in column 3 as in original state, this is also as expected, because the optimized algorithm does not regard the viewport as the whole table), on other lines the viewport contain only aligned rows and `gq1` does nothing.

This is exactly the same algorithm that is used when a table is edited in Insert mode. But imagine that you added a new row in a table (using `Enter` or `Tab`): the last row gets aligned automatically, so now it has many spaces in the columns, and you start adding content: the row gets inevitably longer than the upper rows. Shall the not-optimized fall-back algorithm fire in this case? No, in most cases not! When comparing lengths in the viewport, the current line gets preliminary formatted, i.e. shrinked with `fmt_row()`, so if lengths of data in all columns of the current row are less than lengths of columns in other rows of the viewport, the optimized algorithm triggers.

I see that `gq1`'s behavior is more complex that I thought before. Probably it makes sense to not document it? Or put all the gory details in the doc? Or remove it as a command? (But I would not do so, as it reflects how the optimized algorithm works.)

hq6:
I don't think the gory details belong in the main doc; that should stay high-level and I think the current description is fine.

I think it may be worthwhile to create a new file `DesignNotes` (perhaps `DesignNotes.wiki`?) where we describe these possibly counterintuitive behaviors for posterity. 

Do you think that makes sense? If so, I'll add the stub of a file in the `dev` branch and you can add these notes to that as part of this PR?

Nudin:
> I think it may be worthwhile to create a new file `DesignNotes` (perhaps `DesignNotes.wiki`?) where we describe these possibly counterintuitive behaviors for posterity.
> 
> Do you think that makes sense? If so, I'll add a stub of such a file in the `dev` branch and you can add these notes to that as part of this PR?

I've actually thinking for quite some time now to create a sperate repository `vimwiki/vimwiki-wiki` with a vimwiki in which to collect all tutorials, tips, tricks and background information. I'd propose to add it there.

hq6:
I would agree that such a new repository would be useful to have for users,
but the information associated with this PR feels more like internal
documentation that should be associated with the code in this repository.
That is, this is effectively implementation documentation that is too long
for putting directly into the source file.

Thus, I'd still advocate for putting these particular notes in this main
repository.

Any thoughts on this?

hq6:
It just occurred to me that the question of where this documentation should go is something that can be resolved among maintainers after this PR is merged, because we can always move it later. What's important is that the documentation exists for posterity when this code is committed.

In the interest of expediency, I've gone ahead and created `DesignNotes.wiki`.

@lyokha, Please add the documentation to it and update this PR. We may move it later after @Nudin creates his new repository, but at least it will be there to move.

lyokha:
@hq6, I added description of the newer algorithm in the wiki file. Notice also that

1. The page looks broken on the github, as vimwiki syntax differs from that of github markdown (e.g. the code examples are all skipped). Probably, some *.gitattributes* can help.
2. See *beware* note in the text: the newer algorithm, being much faster for huge tables, is still linear with respect to the height of tables. A probable cause can be syntax highlighting or something else. It would be great if someone test this, as it could be due to unrelated vim settings on my site.

lyokha:
I sorted out the reason for linear slowing down: function `get_indent()` and a new function from another PR `get_cell_aligns()` did not accept *depth* argument and thus were linear according to the table height. I added the depth argument and removed the *beware* note in *DesignNotes.wiki*.

lyokha:
Unfortunately, this new function `get_cell_aligns()` requires traversal to the top of a table in order to find the table separator line and figure out alignments of the columns. Here is the patch to revert to the original implementation of this function:

```diff
diff --git a/autoload/vimwiki/tbl.vim b/autoload/vimwiki/tbl.vim
index 0e67c68..4cc81b7 100644
--- a/autoload/vimwiki/tbl.vim
+++ b/autoload/vimwiki/tbl.vim
@@ -275,9 +275,9 @@ function! s:get_rows(lnum, ...)
 endfunction
 
 
-function! s:get_cell_aligns(lnum, depth)
+function! s:get_cell_aligns(lnum)
   let aligns = {}
-  for [lnum, row] in s:get_rows(a:lnum, a:depth)
+  for [lnum, row] in s:get_rows(a:lnum)
   let found_separator = s:is_separator(row)
     if found_separator
       let cells = vimwiki#tbl#get_cells(row)
@@ -356,7 +356,7 @@ function! s:get_aligned_rows(lnum, col1, col2, depth)
     endfor
     let max_lens = s:get_cell_max_lens(a:lnum, cells, startlnum, rows)
   endif
-  let aligns = s:get_cell_aligns(a:lnum, a:depth)
+  let aligns = s:get_cell_aligns(a:lnum)
   let result = []
   for [lnum, row] in rows
     if s:is_separator(row)
```

This reverts the algorithm to the linear class, but slow-down seems to be not significant even for tables with 1000 and more rows.

Please someone apply this patch or re-open the PR and I will do it.

hq6:
It may be easiest to manage if you create a new PR and mention that it's
fixing a bug introduced in this one.

PR #98, opened 2015-01-08
Use shiftwidth() instead of &shiftwidth

mMontu:
Using &shiftwidth makes VimwikiListChangeLvl misbehave when this option is set
to 0, which is a default configuration of [DetectIndent](https://github.com/roryokane/detectindent).


EinfachToll:
The shiftwidth() function is rather new, but many users, e.g. me, are still running Vim 7.3. Could you make a function like shown in `:h shiftwidth()` and put it in autoload/vimwiki/u.vim and use this instead?

mMontu:
Sure, no problems.

EinfachToll:
Ok, thanks!

PR #103, opened 2015-01-09
Fix misprint in documentation

t7ko:
Fixed a misprint.


t7ko:
Not quite sure since I'm new to github -- should I have pushed this commit directly into `vimwiki/dev`?  Now that I'm a contributor?  pls let me know.

EinfachToll:
Yes, of course you can push trivial fixes like this to dev directly. However, if there is something to discuss, open an issue or a pull request first.

PR #104, opened 2015-01-09
autoload/vimwiki/base.vim was sourced at vim startup

t7ko:
`plugin/vimwiki.vim` was calling a function from `autoload/../base.vim`.  Thus causing the autoload file be loaded at each and every startup, thus defeating the entire idea of that file ("if it's always loaded, why do we need it anyway").

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -8,6 +8,23 @@ if exists("g:loaded_vimwiki_auto") || &cp
 endif
 let g:loaded_vimwiki_auto = 1
 
+" s:vimwiki_get_known_syntaxes
+function! s:vimwiki_get_known_syntaxes() " {{{
+  " Getting all syntaxes that different wikis could have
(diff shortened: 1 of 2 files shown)


EinfachToll:
Good, but
- vimwiki#base#get_known_syntaxes is only called from within base.vim, so there's no need to move this function
- OTOH, if it is only called from within base.vim, we can rename it to s:get_known_syntaxes
- don't call it VimwikiGetKnownExtensions but s:vimwiki_get_known_extensions, so that the function is local to the file. Function names starting with a capital letter are global, but that's unnecessary here and clutters the namespace.

t7ko:
@EinfachToll : fixed.

EinfachToll:
@t7ko: no, only 2 out of my 3 points are fixed ;)

t7ko:
Yes @EinfachToll you're right, I've missed it :)  Now should be fixed.  I hope. :)

PR #106, opened 2015-01-10
Bold anchors were duplicated in get_anchors() results

t7ko:
Open new wiki page, add bold text in (e.g. `*bold*`), and issue:
```
:echo vimwiki#base#get_anchors(@%, 'default')
```

You'll see that `bold` is returned twice.  Apparently it's because `matchstr()` expects that `{count}` argument starts numbering at 1 not at 0 (vim docs say nothing about this, but experiments kind of proved it).

So I've fixed the `vimwiki#base#get_anchors()` to account for it.

(PS: Ubuntu, Vim 7.4)


EinfachToll:
Ok, makes sense.

PR #144, opened 2015-07-07
Fix bug with folds shortening on multibyte characters.

t7ko:
If section title contains multibyte chars, the fold shortening function
often fails, because it works with text as array of bytes, not
characters (see strlen() doc for example).

This commit fixes this issue.

Example header which reproduces the problem for me:
```
====  Установка закладных труб для образования отверстий необходимых для продуха. ====
```


t7ko:
@EinfachToll -- pls review if it's good or not.  Thanks.

EinfachToll:
Works for me and looks good to me :)

Issue #127, opened 2015-04-09
make `g:vimwiki_toc_header` a per-wiki option


embear:
I have multiple wikis in different languages. It would be nice to be able to set the `g:vimwiki_toc_header` different for different wikis to have the correct name in each.

ranebrown:
#257

tinmarino:
Fixed: d75b36620dcf4

PR #152, opened 2015-08-13
change option 'toc_header' from global to local wiki setting

embear:
Hi there,

I just made some small modifications to change option 'toc_header' from global to local wiki setting.
I requested this in issue #127.


EinfachToll:
I'm planning to overhaul the variable handling and generally allow (nearly) all options to be wiki local or global, so I'm closing this PR.

PR #165, opened 2015-10-28
Add command for opening yesterday's diary

Tomsod:
I frequently stay up past midnight, and so when I bother to mark down today's
progress, the date I need to write in diary on is technically yesterday.
There was no quick way to open yesterday's page (VimwikiDiaryPrevDay doesn't
work if yesterday's entry was never created), so I've patched a command.

I think it might prove useful for anyone in the same situation.


EinfachToll:
Thanks, might be useful for some people.

PR #168, opened 2015-11-26
follow url silently in mac

hiberabyss:
Every time follow a link in browser will print `Press ENTER or type command to continue`, this PR will avoid this.


EinfachToll:
Thanks!

Issue #85, opened 2014-11-11
Add support for categories and/or tags (keywords)


t7ko:
I'd like to implement this support.  But I'm new to vimwiki and don't undestand the core philosophy deep enough, so I need some guidance -- how, in general terms, this should be architected.

In particular, the key questions are:
- What should be the syntax?
- What would be the best moment to collect/update list of pages for a given Category/Keyword?

Crazy ideas:
- Add `[[@Category/the name here]]` anywhere in the document.  And then add on-write hook to append to that page.  And then the category page is simply another link.
  - Pros: no new syntax, looks more or less natural, and similar to Wiki.
  - The downside is: subdirectory is treated as a separate wiki, links would need to be put in there with relative path `../link name` to be able to reach upwards.
- Similar to previous: `[[@Category - Categ Name]]`.  Solves the downside above, but pollutes the namespace.

t7ko:
@EinfachToll : I've started working on this feature.  I've got basic stuff ready -- updating metadata file on page save.  Would you like to pre-review it?  It would probably help me.  Alternatively, I can continue and complete entire feature -- but then, it will be more effort to re-do stuff if you find the approach to be wrong.

EinfachToll:
Yes, if you e.g. put it in an extra branch in your private fork, I'll happily take a look.

t7ko:
@EinfachToll : its under name 'tags' ([here](https://github.com/t7ko/vimwiki/tree/tags)).  Yesterday and today, I have implemented pretty much everything the basic version needs.  The only interesting feature that's missing now is search by multiple tags.  Not even sure if it's needed.

What's implemented:
* metadata file support (load, save)
* updating metadata on page save; per-wiki flag to enable/disable it
* tag completion when editing page content
* bulk re-generate metadata for all pages (command)
* docs
* syntax definitions (highlight)
* cmd to search by tag
* cmd to generate wiki page with tags and their "instances"

Notes:
* I'm using very simple format for metadata file (tab-separated list).  It's really easy, 30 lines of code total.  And still provides all that seems to be necessary.
* I'm using "tags" (not "categories") everywhere.

Let me know if you want me to create a pull request out of it.

EinfachToll:
Ok, I tried it and I see that the new commands do something, at least ;)

Some conceptual remarks:
[...]

PR #169, opened 2015-11-26
Fix sorting order of tags.

t7ko:
Before the fix, tags file sorting was done alphabetically.  That would treat line numbers as strings, and so, for example, if the same tag was placed on the same page on lines, say, 9 and 114, the order you would get, 114 would go first, instead of 9.

Fix adds proper entries comparison to the sort function.

### DIFF
--- a/autoload/vimwiki/tags.vim
+++ b/autoload/vimwiki/tags.vim
@@ -213,6 +213,38 @@ function! s:merge_tags(metadata, pagename, file_metadata) "{{{
   return metadata
 endfunction " }}}
 
+" s:tags_entry_cmp
+"   Compares two actual lines from tags file.  Return value is in strcmp style.
+"   See help on sort() -- that's what this function is going to be used for.
+"   See also s:write_tags_metadata below -- that's where we compose these tags
+"   file lines.
+"
+"   This function is needed for tags sorting, since plain sort() compares line
+"   numbers as strings, not integers, and so, for example, tag at line 14
+"   preceeds the same tag on the same page at line 9.  (Because string "14" is
+"   alphabetically 'less than' string "9".)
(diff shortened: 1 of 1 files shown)


t7ko:
@EinfachToll could you pls check whether the fix looks good to you?

The scenario I'm fixing is:

* Enable tags.
* Open any page.
* Put a tag `:test:` on any line in range 3-9, but not below the header (so that tag stays on its own).
* Put another instance of the same tag on any line after 10, but again not below the header.
* Save.
* Run `:tag test`.

**Expected:** jumps to the first instance, at the beginning of file.

**Actual:** jumps to the last one.

t7ko:
Note: this is a bugfix to a feature added in #85.

EinfachToll:
Hm, so actually the tags file will be not sorted, although the first line in the tags file says so. I don't know if that always yields the correct result.
An easier and perhaps safer solution would be to just prepend the line number with zeros before sorting:
~~~viml
      call add(tags,
            \   entry.tagname  . "\t"
            \ . pagename . VimwikiGet('ext') . "\t"
            \ . repeat("0", 6 - strlen(entry.lineno)) . entry.lineno
            \ . ';"'
            \ . "\t" . "vimwiki:" . entry_data
            \)
~~~
It's not that elegant, but works.

t7ko:
As far as I understand, sorting is only required for tags; that is later used to do binary search to locate tag information.  Sorting the entire file is not strictly required.

Anyway.

Would you like me to apply your change and merge it into `dev`?

EinfachToll:
No, we can take your solution. I was only a little concerned about performance at first, but in my quick tests I hardly noticed a slowdown even with a thousand tags.

t7ko:
Well, I'll merge it then.

PR #170, opened 2015-11-27
Bugfix/cursor moves on toc update

t7ko:
I'm frequently using pages with TOC, and I have folds enabled.

With this environment, I've noticed the following:
* cursor moves when you save page right after you add/remove headers (which changes size of TOC section)
* TOC section fold is always closed after save -- even if it was open right before
* screen sometimes scrolls (esp. if you sit within large fold when you save)

This pull request will fix these three issues.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1175,6 +1175,8 @@ function! vimwiki#base#update_listing_in_buffer(strings, start_header,
     let whitespaces_in_first_line = ''
   endif
 
+  let start_of_listing = start_lnum
+
   " write new listing
   let new_header = whitespaces_in_first_line
         \ . substitute(g:vimwiki_rxH1_Template,
@@ -1193,9 +1195,9 @@ function! vimwiki#base#update_listing_in_buffer(strings, start_header,
   endif
 
   " Open fold, if needed
-  if !is_fold_closed
-    exe start_lnum
-    norm zo
+  if !is_fold_closed && foldclosed(start_of_listing)
+    exe start_of_listing
+    norm! zo
   endif
 
   if is_cursor_after_listing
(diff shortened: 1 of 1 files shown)


t7ko:
@EinfachToll -- pls review whether the fix is good enough.

EinfachToll:
> cursor moves when you save page right after you add/remove headers (which changes size of TOC section)

Hah, that bugged me too!

But I get an error in the line

    norm zo

when I have no folds at all. Also, always use `normal!`, not `normal` in scripts, so that it still works if the user has remapped `zo`. So, if you feel like fixing this, that would be cool.

t7ko:
@EinfachToll -- fixed.

autoload/vimwiki/base.vim
```diff
@@ -1171,9 +1191,19 @@ function! vimwiki#base#update_listing_in_buffer(strings, start_header,
   " append an empty line if there is not one
   if start_lnum <= line('$') && getline(start_lnum) !~# '\m^\s*$'
     call append(start_lnum - 1, '')
+    let lines_diff += 1
   endif
 
-  call setpos('.', old_cursor_pos)
+  " Open fold, if needed
+  if !is_fold_closed && foldclosed(start_of_listing)
```
EinfachToll: I still get an error. But it goes away when I write `foldclosed(start_of_listing) > -1` here. I think this is actually what you intended.
With the `> -1`, it looks good in my quick tests, but I don't have that much experience with folds, because I don't use folds in Vim at all. So, if you add it, you can merge everything.

EinfachToll:
@t7ko: BTW something totally unrelated: do you use g:vimwiki_debug or g:VimwikiLog for debugging? If not, I would just remove it entirely.

t7ko:
@EinfachToll -- no, I've never used it.

t7ko:
@EinfachToll -- I've changed the code as you suggested.  Sorry for that stupid bug.  I'm going to merge it in now, but please recheck again on your envriomnent, just to make sure it works as expected.

PR #184, opened 2016-01-10
Include subdir in wiki page name when rebuilding tags.

wangzq:
As title.


EinfachToll:
Thanks. I'm very short on time at the moment, but I'll take a look into this later.

EinfachToll:
I split your changes into two commits and made some changes regarding style, but thanks @wangzq!

PR #186, opened 2016-01-18
fix #64: markdown style link support

### LINKED ISSUE #64: vimwiki is not support markdown style link
Step:
1. change syntax to markdown
2. input enter over a word
result:
[[test]]

Expect the result
[test](test.md)

lotabout:
Now, when normalize a link in markdown syntax, it will get:

```
[some link text](some link text)
[](visual selection text)
```

And the following three links will work( `.md` is the extension specified by `wiki.ext` in the configuration.

```
[[link]]
[link](link)
[link](link.md)
```

### DIFF
--- a/autoload/vimwiki/markdown_base.vim
+++ b/autoload/vimwiki/markdown_base.vim
@@ -108,6 +108,8 @@ function! vimwiki#markdown_base#follow_link(split, ...) "{{{ Parse link at curso
     if lnk != ""
       if !VimwikiLinkHandler(lnk)
         if !vimwiki#markdown_base#open_reflink(lnk)
+          " remove the extension from the filename if exists
+          let lnk = substitute(lnk, VimwikiGet('ext').'$', '', '')
           call vimwiki#base#open_link(cmd, lnk)
         endif
       endif
@@ -173,7 +175,7 @@ function! s:normalize_link_syntax_n() " {{{
   if !empty(lnk)
     let sub = vimwiki#base#normalize_link_helper(lnk,
           \ g:vimwiki_rxWord, '',
-          \ g:vimwiki_WikiLinkTemplate1)
+          \ g:vimwiki_Weblink1Template)
     call vimwiki#base#replacestr_at_cursor('\V'.lnk, sub)
     return
   endif
@@ -192,9 +194,10 @@ function! s:normalize_link_syntax_v() " {{{
   try
     norm! gvy
     let visual_selection = @"
-    let visual_selection = substitute(g:vimwiki_WikiLinkTemplate1, '__LinkUrl__', '\='."'".visual_selection."'", '')
+    let link = substitute(g:vimwiki_Weblink1Template, '__LinkUrl__', '\='."'".visual_selection."'", '')
+    let link = substitute(link, '__LinkDescription__', '\='."''", '')
 
-    call setreg('"', visual_selection, 'v')
+    call setreg('"', link, 'v')
 
     " paste result
     norm! `>pgvd
--- a/autoload/vimwiki/markdown_base.vim
+++ b/autoload/vimwiki/markdown_base.vim
@@ -195,7 +195,7 @@ function! s:normalize_link_syntax_v() " {{{
     norm! gvy
     let visual_selection = @"
     let link = substitute(g:vimwiki_Weblink1Template, '__LinkUrl__', '\='."'".visual_selection."'", '')
-    let link = substitute(link, '__LinkDescription__', '\='."''", '')
+    let link = substitute(link, '__LinkDescription__', '\='."'".visual_selection."'", '')
 
     call setreg('"', link, 'v')
 


robcross:
I'm playing with this change, and it seems to work great for creating and following links in normal mode, but I can't get it working in visual mode. It just exits visual mode when I press enter.

EinfachToll:
@lotabout Thanks, the markdown support definitely needs some improvements. However, I wonder why you chose two different forms of links when pressing enter in normal mode versus pressing enter in visual mode. I think it's less confusing when you get `[text](text)` in both cases.

lotabout:
@EinfachToll In my usage of vimwiki, normally I would use visual selection only when I have to add description for html links. For example:

```
[Google](https://www.google.com)
```
So I think it would be handy to have `[](link)` and then add the description instead of having to delete a bunch of text and add description.

However this is my personal scenario, we can discuss and decide which one is preferred. Thanks!

EinfachToll:
Ok, but then create `[](link)`-style links in both cases, otherwise, it would be inconsistent. Although these links disappear when conceal is on, but we should fix that independently.

lotabout:
@EinfachToll Sorry for late reply. After playing with it for more time, I think your proposal for generate `[text](text)` is more handy.

So I add a new commit to unify the behaviour for visual mode and normal mode.

EinfachToll:
Thanks!

lecagot:
nice to see markdown support going forward!

I do have a question: markdown style interwiki links `[link](link.mkd)` are working now. Pressing `ENTER` on a word now also creates a markdown link `[link](link)`. Any plans to set in the config that vimwiki should also append the markdown extension to the link (pressing `ENTER` --> `[link](link.mkd)`)? this would improve the integration into GitHub and gitit a lot!

looking forward to see markdown style interwiki links support for `VimwikiDiaryGenerateLinks` and `VimwikiGenerateTags` etc. as well (my vimscript skills are limited, should look into that a bit).

EinfachToll:
> I do have a question: markdown style interwiki links [link](link.mkd) are working now. Pressing ENTER on a word now also creates a markdown link [link](link). Any plans to set in the config that vimwiki should also append the markdown extension to the link (pressing ENTER --> [link](link.mkd))? this would improve the integration into GitHub and gitit a lot!

I agree that it is necessary to to include the extension somehow, if you want to use the markdown files elsewhere. I thaught about adding something like a preprocessing step which just adds extensions to all markdown links before you further process them with a markdown coverter. But just allowing the extension in the links and conceal them is probably the better solution.

> looking forward to see markdown style interwiki links support for VimwikiDiaryGenerateLinks and VimwikiGenerateTags etc. as well (my vimscript skills are limited, should look into that a bit).

Yes, given that markdown is wildly used today the markdown support of Vimwiki is not optimal.

lecagot:
here's just a quick hacky python script to convert `[[link]]` vimwiki links to
the markdown syntax `[link](link.mkd)` (it's not beautiful, but it works!).

`links.py`:

```python
#!/usr/bin/env python

import re
import sys

data = []
with open(sys.argv[1], 'r') as f:
    for line in f:
        data += [line.strip('\n')]

extension = 'mkd'

new_data = []
for line in data:
    links_raw = re.findall(r'(\[\[[^(]*?\]\])', line)
    if links_raw != []:
        for li_raw in links_raw:
            li = li_raw.strip('[]')
            if '|' in li:
                pair = li.split('|')
            else:
                pair = [li, li]
            new_link = '[%s](%s.%s)' % (pair[1], pair[0], extension)
            line = line.replace(li_raw, new_link)
	new_data += [line]

with open(sys.argv[1], 'w') as f:
    for line in new_data:
        print(line)
        f.write('%s\n' % line)
```

it can be run for all files in your `vimwiki` folder by

```bash
$ find . -type f -name '*.mkd' -exec links.py {} \;
```

Post/pre-processing to add the extension to interwiki links is also not difficult, but of course it would be more convenient to have it attached when creating the link.

patrickdavey:
I have a [ruby gem](https://github.com/patrickdavey/vimwiki_markdown) for doing basic vimwiki markdown to html conversion. So you can use the `VimwikiAllToHTML` etc. It doesn't support the :local etc.  links (I didn't realise these existed until recently)

By moving from `[[blah]]` style links to straight markdown it make the markdown to html conversion trickier.  You no longer know when something is a vimwiki link or not. The logic now needs to check first as to whether the markdown link is to a local file, or a legitimate link which should not be converted.

Anyway, just a brain dump. I'll update my gem at some point, but thought I'd comment all the same in case I'm missing something!

Thanks so much for Vimwiki - I use it all the time!!

patrickdavey:
I've updated  [my vimwiki_markdown gem](https://github.com/patrickdavey/vimwiki_markdown) . It now supports the various markdown style links, including relative links, absolute links etc.  Still doesn't do the `:local` etc. type links, PR's welcome ;)

PR #187, opened 2016-01-27
Dynamically detect nested syntaxes (fix #128)

### LINKED ISSUE #128: Syntax highlighting for code
It would be nice if vimwiki could syntax highlight code if entered using the ```language syntax just like github does or like https://github.com/xolox/vim-notes is able to do. If for example I write ```php the following code will be highlighted using the vim php syntax. Works really great in vim-notes.

mMontu:
I'm probably missing something on the discussion for #128, but this change doesn't seems inefficient. Maybe it could be disabled for files beyond a certain size, but I think it can be very useful (reducing 80% of my `wiki.nested_syntaxes` entries).

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1926,6 +1926,18 @@ function! vimwiki#base#normalize_link(is_visual_mode) "{{{
   endif
 endfunction "}}}
 
+" vimwiki#base#detect_nested_syntax
+function! vimwiki#base#detect_nested_syntax() "{{{
+  let last_word = '\v.*<(\w+)\s*$'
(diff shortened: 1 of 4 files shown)


EinfachToll:
Ok, you see the new syntax only after exiting and re-entering the buffer, but it's cool nevertheless. Regarding speed/file size, I just would add an option for disabling it. I'll take care of that the next days if you don't want to to it yourself.

mMontu:
I've included it as an wiki option (default on), so it may be disabled for a given wiki that contains huge files.

EinfachToll:
Thanks, @mMontu ! I only did a small cleanup of the doc.

PR #197, opened 2016-03-27
escape paths with spaces when adding to tags

reedcwilson:
The `fnameescape()` method escapes special characters that may be found in a file path. This is needed for file paths containing spaces.

See the [vimdocs](http://vimdoc.sourceforge.net/htmldoc/eval.html#fnameescape()) for more information.


EinfachToll:
I don't think this change does what you expect. Can you please try out the latest version of Vimwiki from the dev branch? I think your problem is already solved.

reedcwilson:
Ah, I wasn't aware of the dev branch. I see the commit: [949b29b](https://github.com/vimwiki/vimwiki/commit/949b29b6943a0448f6d4551d62d2d24843b00496) in dev that works great.

PR #202, opened 2016-04-29
Don't override existing mappings

Ram-Z:
Currently, if a mapping for `<Leader>ww` (for example) already exists, the `nmap` will echo an error message because of the `<unique>` argmuent and not set the mapping. This patch silences this error.

### DIFF
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -309,43 +309,43 @@ command! VimwikiShowVersion call s:get_version()
 
 let s:map_prefix = vimwiki#vars#get_global('map_prefix')
 
-if !hasmapto('<Plug>VimwikiIndex')
+if !hasmapto('<Plug>VimwikiIndex') && maparg(s:map_prefix.'w', 'n') == ""
   exe 'nmap <silent><unique> '.s:map_prefix.'w <Plug>VimwikiIndex'
 endif
 nnoremap <unique><script> <Plug>VimwikiIndex :VimwikiIndex<CR>
 
-if !hasmapto('<Plug>VimwikiTabIndex')
+if !hasmapto('<Plug>VimwikiTabIndex') && maparg(s:map_prefix.'t', 'n') == ""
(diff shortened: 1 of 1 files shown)


Ram-Z:
@EinfachToll Any thoughts on this?

aeosynth:
i'd prefer vimwiki to use  `<LocalLeader>` instead of clobbering or silently failing.

hq6:
@Ram-Z , Thank you for this PR!

While this removes an annoying and cryptic error message, it also makes Vimwiki fail silently if a user has already mapped one of its keys. 

How would you feel about instead outputting a more clear error message when the user has already mapped one of Vimwiki's keybindings? One example would be  "Vimwiki failed to map `<Leader>ww` because it is already mapped, so please map a different key for this command?"

hq6:
In the interest of keeping a clean repository, I am closing this PR due to lack of response from the author.

If there is renewed interest in making errors more clear when a there is a mapping conflict, please feel free to ping on this PR or create a new one.

Ram-Z:
The problem with displaying any message in vim is that it gets in the way of starting vim. This patch was trying to address that issue, printing a different message defeats the point.

I don't think that any user will have troubles debugging this. It's not like it will not do anything at all when the binding is already mapped, it will instead execute the mapping the user has selected (or was provided by a different plugin).

A better approach would be to have a way to disable setting some (or all) mappings.

Ram-Z:
@hq6 Regarding this PR, I'm happy to leave it closed if this solution is not suitable for you, but I'll probably keep the patch alive on my branch as this is essential for me to use vimwiki.

hq6:
This appears to be a reasonable argument for why the user will have trouble debugging this.

@ranebrown @Nudin If you either you also think this PR is a good idea, I'm happy to merge it.

Nudin:
I don't have any problems with this solution even through I'm not sure if it is the best solution.

Ram-Z:
Thanks. :+1:

PR #221, opened 2016-06-22
Allow to use includes in templates

laserb:
This adds the possibility to use includes in the templates by using

```
%include other_template
```

It can be used to share a common head or other common parts among templates.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -111,6 +111,7 @@ function! s:get_html_template(template) "{{{
     let template_name = s:template_full_name(a:template)
     try
       let lines = readfile(template_name)
+      let lines = s:get_html_template_include(lines)
       return lines
     catch /E484/
       echomsg 'Vimwiki: HTML template '.template_name.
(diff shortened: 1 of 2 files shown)


doc/vimwiki.txt
```diff
@@ -1916,6 +1916,11 @@ where
 
   %content% is replaced by a wiki file content.
 
+It is possible to include a template in another template.
+Consider you have a template named 'head.html' that you'd like to have
+in the template 'person.html' then add >
+    %include head.html
```
Nudin: I'm not sure if the format `%include head.html` is the right choise:

- All other formats have the form `%somthing%` this hasn't a defined end.
- There must not be anything after the filename and the filename may not contain spaces.
- include does not say if html or wikitext-file is included.

Maybe `%includehtml:file.html%` would be an alternative? But I'm unsure about this. @johnmarcampbell and @EinfachToll – any thoughts?

autoload/vimwiki/html.vim
```diff
@@ -125,9 +126,25 @@ function! s:get_html_template(template) "{{{
...
   return lines
 endfunction "}}}
 
+function! s:get_html_template_include(lines) "{{{
+  let html_lines=[]
+  for line in a:lines
+    if line =~# '%include'
+      let template = []
+      let template = split(line)
+      let include_lines = s:get_html_template(template[-1])
```
Nudin: Don't simply use the last word of the line but the one after the `%include` (or different markup, see below).

johnmarcampbell:
I think @Nudin's makes good points above. The modified syntax (`%includehtml:file.html%`) looks good to me.

hq6:
I am closing this PR as stale, since the original author has not responded to review comments in over a year.

Issue #225, opened 2016-07-08
Delay to edit long files


mMontu:
I'm experiencing large delays when editing at the _end_ of large files. Using `:substitute` causes my machine to freeze. Here is a procedure to reproduce it:

```
:e temp.wiki
i<cr>abc<cr>def<esc>
kA<s-cr>{{{<cr>}}}<esc>k
:r $VIMRUNTIME/doc/index.txt
zc

This leads to something like this:

         - abc
           {{{ [1636] *index.txt*     For Vim version 7.4
         - def

:%y
G5p
G
o   --> long delay to open the new item
ghi --> long delay to inset the text
2GA<cr> def --> no delay
```

I noticed that even with no folds it is still  slow. 

I imagine that the problem witch `<cr>` might occurs because the plugin read all precedent lines in order to understand what is the correct indentation/list bullet to insert.
But what could cause the delay on regular keys (`ghi`)? Maybe the delay in regular insert is the most relevant factor.

EinfachToll:
The reason in both cases is the folding. If you disable the folding (i.e. set `g:vimwiki_folding=''` in the vimrc, not just open all folds) it will probably work. Maybe the folding code can be a bit optimized, but not much, I'm afraid.

mMontu:
Disabling folding indeed solved the problem. I was using folding `'expr'`, which I understood is the fastest method according to the docs. I was wondering if it is possible to use a different marker or an extra identifier for long folds in order to make it faster.

ranebrown:
There have been some improvements to fold handling such as the `quick` option and #601. Other fixes or enhancements should be opened as separate PRs/issues.

infinitewhileloop:
I was experiencing the same issue. If anyone else wants to keep folding enabled, but avoid the slowdown, I found a solution here: https://github.com/vimwiki/vimwiki/issues/461#issuecomment-463249175

The trick is to use vim-pandoc, which folds Markdown files more efficiently.

```vim
Plug 'vim-pandoc/vim-pandoc'
Plug 'vim-pandoc/vim-pandoc-syntax'

let g:vimwiki_folding='expr'
let g:pandoc#filetypes#handled = ["pandoc", "markdown"]
let g:pandoc#filetypes#pandoc_markdown = 0
let g:pandoc#folding#mode = ["syntax"]
let g:pandoc#modules#enabled = ["formatting", "folding"]
let g:pandoc#formatting#mode = "h"

augroup vimwiki_folding
  autocmd!
[...]

PR #227, opened 2016-07-12
Workaround for slow folding methods (issue #225)

mMontu:
It seems that Vim has a limitation when folding by expression and syntax. The Vim FAQ 29.7 mentions a [workaround on the wiki ](http://vim.wikia.com/wiki/Keep_folds_closed_while_inserting_text) which seems to work fine. 

The creation of the new lists still has a (smaller) delay, probably because the `<CR>` mapping leaves insert mode.


EinfachToll:
Even including updated documentation! Very good!

Issue #236, opened 2016-08-28
highlight in math environment


ycpei:
This is both a issue report and my-attempt-to-fix-the-issue report:

When I try to use the maths environment, say

```
{{$
a^b
}}$
```

it is highlighted as if I am using maths in a non-maths environment, that is, the `^` is red.
However, if I put 

```
{{$
$a^b
}}$
```

the red error-highlighting is gone and the content after `$` in the second line is treated as maths in latex, but after conversion to html there would be an extra `$` at the beginning of the equation.

I am trying to solve this problem by delving into the syntax files of both vimwiki and latex (`tex.vim`)

These lines 608-613 in `syntax/vimwiki.vim` caught my attention:

```
" LaTeX
call vimwiki#base#nested_syntax('tex', 
      \ g:vimwiki_rxMathStart.'\%(.*[[:blank:][:punct:]]\)\?'.
      \ '\%([[:blank:][:punct:]].*\)\?',
      \ g:vimwiki_rxMathEnd, 'VimwikiMath')
"}}}
```

as well as line 509

```
hi def link VimwikiMath Number
```

I am not quite familiar with vimscript, but some googling tells me this line links `VimwikiMath` group to `Number` group.

But how is the `Number` group related to the syntax highlight in `tex.vim`?

Inspecting `tex.vim`, I notice the following lines

```
syn region texMathZoneV matchgroup=Delimiter start="\\("            matchgroup=Delimiter    end="\\)\|%stopzone\>"          keepend concealends contains=@texMathZoneGroup
syn region texMathZoneW matchgroup=Delimiter start="\\\["           matchgroup=Delimiter    end="\\]\|%stopzone\>"          keepend concealends contains=@texMathZoneGroup
```

So I added 

```
syn region texMathZoneU matchgroup=Delimiter start="{{\$"           matchgroup=Delimiter    end="}}\$\|%stopzone\>"         keepend concealends contains=@texMathZoneGroup
```

I also changed
    syn cluster texMathZones        contains=texMathZoneV,texMathZoneW,texMathZoneX,texMathZoneY,texMathZoneZ

to

```
syn cluster texMathZones        contains=texMathZoneU,texMathZoneV,texMathZoneW,texMathZoneX,texMathZoneY,texMathZoneZ
```

and added

```
HiLink texMathZoneU     texMath
```

Now the highlight of 

```
{{$
a^b
}}$
```

works as expected in `.tex` files, but the problem in `.wiki` files persists...

[...]

PR #242, opened 2016-09-07
Minor walkaround to a problem in math block syntax highlight (Issue #236)

ycpei:
[Issue 236](https://github.com/vimwiki/vimwiki/issues/236) is the same as the problem described in [this stackoverflow post](http://stackoverflow.com/questions/13611303/vim-maths-syntax-highlighting). Hence the solution is the same.

Note it is not a perfect fix because say in 

```
{{$
a^b
}}$
```

the a and b are marked as usual text, rather than maths, which should be HiLink-ed as `SpecialComment`.

But at least the red error-highlight of `^` is gone.


EinfachToll:
I did not merge you PR, because it would cause **every** TeX snippet to be highlighted like math (particularly preformatted text), so I changed your patch a little. Good work, though!

PR #247, opened 2016-10-08
Added the ability to turn off mapping.

Mephistophiles:
Added option for enable/disable mapping (default: enable)

For disable let g:vimwiki_use_mapping = 0

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -2316,6 +2316,19 @@ Default: 1
 
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


EinfachToll:
What's wrong with the already existing option g:vimwiki_map_prefix?

johnmarcampbell:
It looks like the conversation here has died out, and as @EinfachToll there is already a way to do this with existing functionality. I'm closing this PR without merging. If anyone wants this PR to be merged we can reopen and discuss it.

Issue #255, opened 2016-11-22
Move mapping for ENTER to ALT-ENTER?


philiprhoades:
People,

I have started using VimWiki - which suits me in that it allows me to move my collection of *.txt files to their *.wiki equivalents but I _can't_ get used to using ENTER for going to another link (file) - I use ENTER _constantly_ for going to the beginning of the next line - is it possible to move the VimiWiki fn for ENTER to ALT-ENTER?

Thanks,
Phil.

EinfachToll:
See `:h vimwiki_<CR>` for how to remap Enter. Putting `:nmap <A-CR> <Plug>VimwikiFollowLink` in your .vimrc would be the right way, but it depends on your terminal if you actually can use Alt+Enter as a key in Vim. At least for me, it doesn't work.

philiprhoades:
No, doesn't work for me either:

Fedora 25 x86_64
xfce4-terminal v0.8.0

Any ideas about where to look to fix it?

Thanks!

philiprhoades:
I am using <C-A> for the time being - I tried using logkeys to work out what was going on but didn't get anywhere . .

EinfachToll:
Maybe it works in Neovim. AFAIK it was one of their goals to make alt keys work properly. Other than that, I know of no better solution than using a different key.

christianrondeau:
I'm using `<CR>` as `<ESC>` by default, and vimwiki takes over. A simple setting to avoid remapping all `<CR>` mapping variations would solve everything. I'm ready to make a pull request, if you think it makes sense.  Something like `g:vimwiki_prevent_cr_remap = 1`

christianrondeau:
To disable all cr mappings :

    nnoremap łwf <Plug>VimwikiFollowLink
    nnoremap łws <Plug>VimwikiSplitLink
    nnoremap łwv <Plug>VimwikiVSplitLink
    nnoremap łwv <Plug>VimwikiTabnewLink
    vnoremap łwn <Plug>VimwikiNormalizeLinkVisualCR
    inoremap łwn VimwikiReturn

calsioro:
This might be related, should probably be a separate issue:
When giving a count to `i` `I` `a` or `A`, the inserted text repeats after pressing enter OR escape, but it repeats in the same line either way.

To reproduce:
- open a new vimwiki buffer, and in normal mode enter 10 and one of {`i`, `I`, `a`, `A`}, type "abc" press `<CR>`.

[...]

PR #262, opened 2016-12-03
255: Disable <CR> maps w/ g:vimwiki_prevent_cr_remap

christianrondeau:
For #255,  allows not mapping enter by setting:

    let g:vimwiki_prevent_cr_remap = 1

In your vimrc. 

The same principle could be applied to other keys, like tab,  arrows,  etc.

### DIFF
--- a/ftplugin/vimwiki.vim
+++ b/ftplugin/vimwiki.vim
@@ -333,6 +333,7 @@ if g:vimwiki_use_mouse
   nnoremap <silent><buffer> <RightMouse><LeftMouse> :VimwikiGoBackLink<CR>
 endif
 
+let s:vimwiki_override_cr = get(g:, 'vimwiki_prevent_cr_remap', '0') != 1
 
 if !hasmapto('<Plug>Vimwiki2HTML')
   exe 'nmap <buffer> '.g:vimwiki_map_prefix.'h <Plug>Vimwiki2HTML'
@@ -346,19 +347,19 @@ endif
 nnoremap <script><buffer>
       \ <Plug>Vimwiki2HTMLBrowse :Vimwiki2HTMLBrowse<CR>
 
-if !hasmapto('<Plug>VimwikiFollowLink')
+if !hasmapto('<Plug>VimwikiFollowLink') && s:vimwiki_override_cr
   nmap <silent><buffer> <CR> <Plug>VimwikiFollowLink
 endif
 nnoremap <silent><script><buffer>
       \ <Plug>VimwikiFollowLink :VimwikiFollowLink<CR>
 
-if !hasmapto('<Plug>VimwikiSplitLink')
+if !hasmapto('<Plug>VimwikiSplitLink') && s:vimwiki_override_cr
   nmap <silent><buffer> <S-CR> <Plug>VimwikiSplitLink
 endif
 nnoremap <silent><script><buffer>
       \ <Plug>VimwikiSplitLink :VimwikiSplitLink<CR>
 
-if !hasmapto('<Plug>VimwikiVSplitLink')
+if !hasmapto('<Plug>VimwikiVSplitLink') && s:vimwiki_override_cr
   nmap <silent><buffer> <C-CR> <Plug>VimwikiVSplitLink
 endif
 nnoremap <silent><script><buffer>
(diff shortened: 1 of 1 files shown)


christianrondeau:
After thinking about this, a much cleaner way to go would be to define a dictionary of all shortcuts, and enable / disable / override them individually. If you agree, do not take this pull request and I'll do another.

christianrondeau:
I'll close this one... Let me know if it's worth doing though (i.e. Will you take the pr)

christianrondeau:
I began looking at this, and the best course of action would be to:
1. Create a dictionary of all mappings (key = function, value = mapping string) 
2. Iterate through all mappings and create them

This is a pretty big refactoring, I'm not sure you'll like it :P

The alternative is to just concatenate some mappings that could be problematic, like enter, using `exe`b like you did elsewhere. Simpler, but kind of a partial solution.

Any thoughts

EinfachToll:
What do you want to achieve in the first place? As you can see in the lines you changed in your PR, the keys are not mapped if the user already has set a mapping to the command in the .vimrc. And if you really don't want a key for a certain command, you can map a non-existing “dummy” key, like `<F13>`.

christianrondeau:
@EinfachToll thanks for answering :) I agree I can do that for *some* mappings (I mapped them to some random unused key, like the "yen" sign), but cases like this:

```
if maparg('<CR>', 'i') !~? '<Esc>:VimwikiReturn'
  inoremap <silent><buffer> <CR> <Esc>:VimwikiReturn 1 5<CR>
endif
if maparg('<S-CR>', 'i') !~? '<Esc>:VimwikiReturn'
  inoremap <silent><buffer> <S-CR> <Esc>:VimwikiReturn 2 2<CR>
endif
```

are a little bit more problematic, since they take over as soon as they are not mapped (and the one I *don't* want mapped is `<CR>` in insert mode)

I thought, if I'm to apply two solutions (a fake mapping, and a PR for the insert mode's `<CR>`), I should probably do something cleaner and easier (such as a list of mappings that can easily be overriden), but it'll be a pretty big PR in terms of code changes (the code itself would not be that complex however).

Something like:

```
let g:vimwiki_mappings = {
  \ 'SomeFunction`: [ 'i', '\<cr>' ],
}
```

and

```
if hasmap(...) && vimwiki_mappings['SomeFunction']
  exe vimwiki_mappings['SomeFunction'][0] . 'noremap <silent><buffer> ' . vimwiki_mappings['SomeFunction'][1] . ' :call SomeFunction()'
endif
```

My only goal though is simply to prevent vimwiki from taking over my `<CR>` mapping in insert mode :)

EinfachToll:
OK, I agree that Vimwiki should not simply map `<CR>` without the possibility for the user to prevent it, but it should suffice to just put a `if !hasmapto('VimwikiReturn', 'i') ... endif` around the lines you quoted, wouldn't it?

christianrondeau:
I'm not sure I can do that. I'd have to check for a mapping of `<Esc>:VimwikiReturn 1 5<CR>` for example. I'm not sure if I'll have to actually convert these to `<Plug>VimwikiReturn` for `hasmap` to work... This will still change the code quite a bit. And I'm not sure if there is a reason for this particular check before the mapping. So in short, I guess I could figure out a way to check whether the function has been mapped somehow, but this feels hacky. 

I do understand that introducing big code changes is not desirable, which is why I chose the global flag in the first place even though it's also hacky. Let me know what you think, worst case I simply fork and the problem is solved for me :)

EinfachToll:
> I'd have to check for a mapping of <Esc>:VimwikiReturn 1 5<CR> for example. 

Why? `hasmapto('VimwikiReturn', 'i')` checks if there is **any** mapping to this command, whether it's `:VimwikiReturn 1 5` or `:VimwikiReturn 2 2`. So, if you, for example put `inoremap <silent><buffer> <F13> <Esc>:VimwikiReturn` into your vimrc, no mapping for `i_<CR>` will be created. Isn't this what you want?

christianrondeau:
Looks like you are right, I didn't know `hasmapto` accepted partial mappings. All right then, if you don't support the idea of the mappings dictionary, I'll close this PR and simply use fake mappings. 

Thanks for taking the time to consider this PR, I'll suggest this in #255 for the next guy!

PR #264, opened 2016-12-05
Prevent imap of `<CR>` if already mapped

christianrondeau:
Following up on #255 and #262, just a check for the imap of cr


j718:
How do I use this prevent vimwiki from taking over <CR> so far I have just deleted the following lines.
```  if maparg('<CR>', 'i') !~? '<Esc>:VimwikiReturn'
    inoremap <silent><buffer> <CR> <Esc>:VimwikiReturn 1 5<CR>```

Is there a way to do this in my vimrc?

PR #266, opened 2016-12-12
Fix typo in docs

NullInfinity:
Correct _arbitray_ to _arbitrary_.


EinfachToll:
After a long and careful review, this LGTM ;)

PR #270, opened 2016-12-12
Fix example mapping for decreasing list item level

priomsrb:


EinfachToll:
Thanks, it makes more sense now.

PR #273, opened 2016-12-15
fixed #272

### LINKED ISSUE #272: Not support markdown header syntax by symbol "-" and "=" underline
markdown HEADER1 and HEADER2 has another format
eg:

This is Header1
========

This is Header2
--------

icersong:
fixed #272


icersong:
Markdown syntax hilight for header with ---- and ====

EinfachToll:
Uh, what happend to this PR? Why does it contain so much stuff which is already in the code? Any chance to rebuild the PR?

hq6:
There have been no updates since EinfachToll's last comment, so I am closing this PR due to staleness.

Nudin:
Since the "real content" of this PR is just two lines and it would apparently fix an often asked issue it might be worth trying out if that works and has no negative side-effects and if so make a new PR.

PR #277, opened 2016-12-18
Open wiki in a split pane

danetrata:
Opening a wiki in a split pane added to function base#goto_index() through the 'tabedit' argument.
The function can now take the following:
```
vimwiki#base#goto_index(..., [1, 2, 3])
1       tabedit
2       split
3       vsplit
```

The change does not affect any other function calls because all the other function calls assumed '1' to be 'tabedit'.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1301,8 +1301,12 @@ function! vimwiki#base#goto_index(wnum, ...) "{{{
     let idx = 0
   endif
 
-  if a:0
+  if a:0 == 1
     let cmd = 'tabedit'
+  elseif a:0 == 2
+    let cmd = 'sp'
+  elseif a:0 == 3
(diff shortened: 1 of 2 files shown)


Nudin:
@danetrata: Thanks for contributing! And sorry for the long delay.
Can you please add consistent behavior to the diary (`vimwiki#diary#make_note` in diary.vim) – then I will happily merge the patch.

danetrata:
Thanks for looking at the PR! 
I just wanted to leave a note how I use it.

`    nnoremap <Leader>wp :call vimwiki#base#goto_index(v:count, 3)<cr>
`

Nudin:
Merging now. Not sure if/how we could document this…

EinfachToll:
@Nudin we need to introduce new commands similar to VimwikiIndex and VimwikiTabIndex and document these commands.

Issue #275, opened 2016-12-17
Allow g:vimwiki_listsyms be larger than 5 elements


Nudin:
Allowing g:vimwiki_listsyms to be an arbitrary size would allow smoother transitions for the parents like:
▁▂▃▄▅▆▇▉

If it contains n elements one would use element 0 for progress == 0, element n-1 for progress == 100 and element ceil(progress/100*(n-1)) for all other values of progress.

EinfachToll:
Thanks!

PR #279, opened 2016-12-21
Allow g:vimwiki_listsyms be larger than 5 elements (Issue #275)

Nudin:
Allows g:vimwiki_listsyms to be of arbitrary size, so that smoother transitions like '▁▂▃▄▅▆▇▉' or '○◔◑◕●✓' are possible.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -861,11 +861,11 @@ function! s:process_tag_list(line, lists) "{{{
     if !empty(chk) && len(chk[1]) > 0
       let completion = index(g:vimwiki_listsyms_list, chk[1])
       let n = len(g:vimwiki_listsyms_list)
-      if completion >= 0 && completion < n-1
-        let completion = float2nr(ceil(completion * 4.0 / (n-1) ))
+      if completion == 0
+        let st_tag = '<li class="done0">'
+      elseif completion > 0 && completion < n
+        let completion = float2nr(round(completion / (n-1.0) * 3.0 + 0.5 ))
         let st_tag = '<li class="done'.completion.'">'
-      elseif completion == n-1
-        let st_tag = '<li class="done4">'
       endif
     endif
     return [st_tag, '']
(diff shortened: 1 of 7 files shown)


EinfachToll:
This is cool in principle, but there are a few more places in the source code which have to be adjusted. If you grep for `listsyms` in the source code you can easily fix it, so that the highlight works again.

Nudin:
Thanks for the feedback. I've now hopefully adjusted all occurrences of `listsyms `. For the HTML-Version I so far translate back to the 5-step-system since I couldn't figure out a possibility to support arbitrary number of symbols there.

Nudin:
Any further opinions to this Pull request?

EinfachToll:
I forgot to hit the *Submit review* button. There are still some issues with your patch.

autoload/vimwiki/html.vim
```diff
@@ -860,8 +860,12 @@ function! s:process_tag_list(line, lists) "{{{
     let chk = matchlist(a:line, a:rx_list)
     if !empty(chk) && len(chk[1]) > 0
       let completion = index(g:vimwiki_listsyms_list, chk[1])
-      if completion >= 0 && completion <=4
+      let n = len(g:vimwiki_listsyms_list)
+      if completion >= 0 && completion < n-1
+        let completion = float2nr(ceil(completion * 4.0 / (n-1) ))
```
Nudin: Is fixed now.

autoload/vimwiki/html.vim
```diff
@@ -861,11 +861,11 @@ function! s:process_tag_list(line, lists) "{{{
     if !empty(chk) && len(chk[1]) > 0
       let completion = index(g:vimwiki_listsyms_list, chk[1])
       let n = len(g:vimwiki_listsyms_list)
-      if completion >= 0 && completion < n-1
-        let completion = float2nr(ceil(completion * 4.0 / (n-1) ))
+      if completion == 0
+        let st_tag = '<li class="done0">'
+      elseif completion > 0 && completion < n
+        let completion = float2nr(round(completion / (n-1.0) * 3.0 + 0.5 ))
```
Nudin: Here you can see the function:
http://www.wolframalpha.com/input/?i=plot+round(x*3%2B0.5)+for+x+in+%5B0,1%5D

PR #286, opened 2017-01-08
Better link creation behavior + removal of problematic string literal construction in substitutions

benburrill:
The string literal thing is related to the links because it resulted in one of the issues I encountered with them.  Due to the wide usage of the pattern that resulted in the bug I fixed it everywhere, not just link-related stuff.

I also removed the restriction that links can only be created in characterwise-visual mode.

Testing has been minimal, mostly just a verification that `<CR>` in linewise-visual and blockwise-visual modes selecting something containing a `'` works as expected.  I may try to play around with some of the places that the string literal change touched later, but I'm fairly confident there won't be any big problems.


benburrill:
The main problem this was meant to solve has since been fixed by #411.  I would argue that my solution is better since it uses the builtin features of `substitute()` rather than defining a new function (twice!), but it doesn't really matter, both are correct.

The other part of my PR (which probably should have been a separate PR in retrospect) extended link creation to all visual modes that stay on one line (not just `v`).  This is convenient for `V` because sometimes I want to turn an entire line into a link.  If you are interested in this feature, I would be happy to create a new PR for it without the now-mostly-obsolete fix for `substitute()` misuse.

Issue #200, opened 2016-04-20
markdown syntax for TOC


lecagot:
so far, `VimwikiTOC` generates a table of contents in the vimwiki syntax, even if markdown syntax is specified:

```
# Contents
    - [[#one|one]]
    - [[#two|two]]
    - [[#three|three]]
    - [[#four|four]]
    - [[#five|five]]
```

any plans on implementing a markdown syntax TOC?

assaf758:
in the meantime markdown TOC can be created with this plugin:
https://github.com/mzlogin/vim-markdown-toc

llinfeng:
@EinfachToll May I revisit this issue and ask: **can I specify the Github style for TOC**? I prefer `:VimwikiTOC` over the [`:GenTocGFM`](https://github.com/mzlogin/vim-markdown-toc) alternative, as the latter does not generate any numbering for the TOC entries.

ANGkeith:
Hey, you can add this configuration for your `mzlogin/vim-markdown-toc` configurations in the vimrc

`let g:vmt_list_item_char = '1.'` so that the toc will be numbered if that is what you want

llinfeng:
@ANGkeith I got a list of TOC with all items enumerated as `1.`  :( I think the documentation said the following:
```
7. `g:vmt_list_item_char`
    default: `*`
    The list item marker, it can be `*`, `-` or `+`.
```

ANGkeith:
@llinfeng Hey, yeah its the markdown syntax for [ordered listing](https://guides.github.com/features/mastering-markdown/). If you read it on github, it will show the ordered list as you wished

```
1. Item 1
1. Item 2
1. Item 3
   1. Item 3a
   1. Item 3b
```

llinfeng:
I see I see. This is helpful! Thanks a lot!

PR #296, opened 2017-02-07
Add config option to allow use Markdown toc format

mzlogin:
ref #200


mzlogin:
Introducing a new option is for compatibility.

If you consider it's OK to make it the standard way, I will create a new PR for it.

llinfeng:
May I ask what is the config option exactly? I am looking forward to a combination of the following two features from the `:VimwikiTOC` command:
1. Github-compatible links (now, they are not compatible)
2. Enumerated list of links in the section named `# Contents` (this is already available).

@mzlogin Just to comment on your [`vim-markdown-toc`](https://github.com/mzlogin/vim-markdown-toc) plugin ==> wonderful job, in particular for handling the good number of different flavors of the Markdown-links. For now, I did not find an option to generate an Enumerated TOC. Otherwise, I shall happily stick with your plugin.

Issue #117, opened 2015-03-05
Can't fold list items


asldjkv:
With [vimwiki](https://github.com/vim-scripts/vimwiki) you can fold headers and lists as following:

Example for headers and list folding:

```
= My current task =
  * [ ] Do stuff 1
    * [ ] Do substuff 1.1
    * [ ] Do substuff 1.2
      * [ ] Do substuff 1.2.1
      * [ ] Do substuff 1.2.2
    * [ ] Do substuff 1.3
  * [ ] Do stuff 2
  * [ ] Do stuff 3
```

Hit <kbd>zM</kbd>:

```
= My current task = [8] --------------------------------------
```

Hit <kbd>zr</kbd>:

```
= My current task =
  * [ ] Do stuff 1 [5] --------------------------------------
  * [ ] Do stuff 2
  * [ ] Do stuff 3
```

Hit <kbd>zr</kbd> one more time:

```
= My current task =
  * [ ] Do stuff 1
    * [ ] Do substuff 1.1
    * [ ] Do substuff 1.2 [2] -------------------------------
  * [ ] Do substuff 1.3
  * [ ] Do stuff 2
  * [ ] Do stuff 3
```

The header-folding does works for me, but list-folding as underneath doesn't:

```
= My current task =
  * [ ] Do stuff 1 [5] --------------------------------------
  * [ ] Do stuff 2
  * [ ] Do stuff 3
```

Vimwiki says this about list-folding, but I don't understand what it exaclty means:

> NOTE: If you use the default vimwiki syntax, folding on list items will work
> properly only if all of them are indented using current shiftwidth.
> For MediaWiki, `*` or `#` should be in the first column.

My vimrc is as following:

```
Plugin 'vimwiki'
set nocompatible
filetype plugin on
syntax on
set foldmethod=syntax
```

asldjkv:
Update: 

I've added the following to my .vimrc

    let g:vimwiki_folding = 'list'

This had the result that headers and list couldn't be folded.

EinfachToll:
That you can only fold headers as you describe in your first post, can be explained by the line

    set foldmethod=syntax
in your vimrc. It lets Vim (not Vimwiki) figure out foldable regions based on the syntax.

The line you added as you wrote in your second post should normally cause Vim to fold list items. For me, it works. What is your output for `:set foldmethod?` ?

[...]

PR #297, opened 2017-02-09
Update folding documentation to match current behaviour.

pjf:
As mentioned in #117, it's not possible to have both section and list
folding working at the same time, but the previous documentation suggested
this was possible in the example.

This change makes the documentation example consistent with current
behaviour.


EinfachToll:
I made a few minor changes to your patch and included it. Thanks!

Issue #308, opened 2017-03-01
Increase/Decrease Todo-List status


Nudin:
With :VimwikiToggleListItem we can change the status of a listitem between 0% and 100% – but we also include sub states configured in g:vimwiki_listsyms – for now we use them mostly to indicate the status of a item with subitems but it can already be nested, so it is possible to use the inbetween-states also for items without subitems.
This would become easier and more obvious if there are functions and keybindings to increment and decrement the state.

I've written a quick-and-dirty patch [in my branch](https://github.com/Nudin/vimwiki/commit/5eb6b1c2ab8c0417a61a8d5d3976ad2b99c65556).
I'll do some code-reuse and documentation but maybe you can already tell me what you think. Also I'm not sure which keybindings we should use by default…

johnmarcampbell:
It looks like there's a merged PR that solves this, so I'm closing the issue.

PR #313, opened 2017-03-08
Possibility to increase/decrease checkbox

Nudin:
This implements Issue #308:
With :VimwikiToggleListItem we can change the status of a listitem between 0% and 100% – but we also include sub states configured in g:vimwiki_listsyms – for now we use them mostly to indicate the status of a item with subitems but it can already be nested, so it is possible to use the inbetween-states also for items without subitems.
This would become easier and more obvious if there are functions and keybindings to increment and decrement the state.

Adds new functions: `vimwiki#lst#increment_cb` & `vimwiki#lst#decrement_cb`
Adds keybinding for them: `<ctrl>-<alt>-<+>` &  `<ctrl>-<alt>-<->`

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ A Personal Wiki For Vim
 ==============================================================================
 
 ![screenshot1](doc/screenshot_1.png)
-![screenshot2](doc/screenshot_2.png)
+![screenshot2](doc/screenshot_2.png) *
 
 Intro
 ------------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


EinfachToll:
In principle, this looks good to me, but my guess is that there are so few people who will use this feature that we should not waste the keys M-+ and M-- for it. Since we already have kind of reserved key combos starting with gl for list manipulation, I suggest gl+ and gl- instead. Can you change that?

Nudin:
Great to see you back in action and thanks for the feedback! I agree that M-+ and M-- aren't ideal and the user can remap it to what we likes anyway – so I can change it to gl+ and gl- instead.

I'm right now setting up my Vimwiki-dev-dir again after accidentally deleting it and screwing the gittree up. ;) Will rebase and patch as soon as this is done…

Nudin:
I totally screwed the rebasing up :( – so I deleted the branch and reapplied the patches. Sry for the inconvenience…
Opened a new pull request: #369.

PR #321, opened 2017-03-29
Add option for enabling local folding settings

augustus-rex:
Because the default folding setting set 'foldmethod=manual' it
makes it difficult to use local folding settings defined in the
after-directory. This code adds a 'local' option to g:vimwiki_folding
that makes this easier.

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -2321,6 +2321,7 @@ Value           Description~
 'expr'          Folding based on expression (folds sections and code blocks)
 'syntax'        Folding based on syntax (folds sections; slower than 'expr')
 'list'          Folding based on expression (folds list subitems; much slower)
+'local'		Enable use of local folding settings
 
 Default: ''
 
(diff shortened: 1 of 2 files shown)


EinfachToll:
I didn't merge your PR because I gave the option a different name (for me, it was not really clear what 'local' is supposed to mean) and pushed it, but thanks for the inspiration!

augustus-rex:
Glad you liked it! And I agree, your changes were clearly better.

A little thought though. In the future when people do stuff like this, come with a good idea that is slightly off in execution, it might be a good idea to encourage them to change it and then merge their pull request instead of just creating a new one. Having your own pull request merged brings a greater sense of accomplishment, which in turn can help to encourage further contributions. Also it can be a way to help people learn :)

Also, thanks for a great plugin. I really do like it ^^

augustus-rex:
If you're interested, I can report that the changes are working splendidly

EinfachToll:
> A little thought though. In the future when people do stuff like this, come with a good idea that is slightly off in execution, it might be a good idea to encourage them to change it and then merge their pull request instead of just creating a new one. Having your own pull request merged brings a greater sense of accomplishment, which in turn can help to encourage further contributions. Also it can be a way to help people learn :)

You know, there are different kinds of people and it isn't so seldom that somebody makes a contribution, I ask for a change and never hear again from this person. Anyway, I will do what you suggested.

PR #326, opened 2017-04-04
Fix issue in Markdown visual mode link creation

### LINKED ISSUE #325: Visual mode Markdown link creation
Selecting more than one word and pressing <CR> doesn't create a link.

salazar:
Fixes #325

### DIFF
--- a/autoload/vimwiki/markdown_base.vim
+++ b/autoload/vimwiki/markdown_base.vim
@@ -145,7 +145,7 @@ function! s:normalize_link_syntax_v() " {{{
     call setreg('"', link, 'v')
 
     " paste result
-    norm! `>pgvd
+    norm! `>""pgvd
 
   finally
     call setreg('"', rv, rt)


EinfachToll:
I merged it because the patch looks harmless and you are not the only one with the issue. But I can't reproduce it and looking at the existing code I can't see what could go wrong. `p` per default pastes from the `"` register, doesn't it? Can you give more detailed steps for how to reproduce it?

salazar:
In other formats — e.g., Vimwiki's own syntax — links can be constructed from more than one work by visually selecting them and pressing Enter. In Markdown, without this change, that won't work. (IIRC, I looked up how the substitution works for Vimwiki's own syntax and spotted that small difference.)

EinfachToll:
Strange, it worked for me even without your patch. And I have no idea why it would not work for everybody else. Anyway, Vim is full of small secrets ;)

PR #331, opened 2017-04-12
add option to convert multiple rows in table to single HTML table row

standardeyre:
This pull request adds a toggle `<leader>wz`.

The toggle's default setting (zero) gives each row its own cell when
converted to HTML. This is what Vimwiki normally does.

If the toggle is set to one, each vimwiki table row is put into the same
single cell when converted to HTML.

Minimum Working Example:

This vimwiki text:

    | One again, you enter a  |
    | place filled with       |
    | twisty little passages, |
    | all alike.              |
    |                         |
    | There's no food.        |

produces this HTML by default after  `:Vimwiki2HTML`:

    <table>
    <tr>
    <td>
    One again, you enter a
    </td>
    </tr>
    <tr>
    <td>
    place filled with
    </td>
    </tr>
    <tr>
    <td>
    twisty little passages,
    </td>
    </tr>
    <tr>
    <td>
    all alike.
    </td>
    </tr>
    <tr>
    <td>
    <br /><br />
    </td>
    </tr>
    <tr>
    <td>
    There's no food.
    </td>
    </tr>
    </table>

With this fork to vimwiki, I get the HTML I want when  `<leader>wz` is set
to one and I then run `:Vimwiki2HTML`:

    <table>
    <tbody><tr><td>
    One again, you enter a
    place filled with
    twisty little passages,
    all alike.
    <br><br>
    There's no food.
    </td></tr>
    </tbody></table>

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -7,6 +7,8 @@ if exists("g:loaded_vimwiki_auto") || &cp
   finish
 endif
 let g:loaded_vimwiki_auto = 1
+" by default, one column has multiple rows
+let g:vimwiki_one_column_one_row = 0 
 
 " s:vimwiki_get_known_syntaxes
 function! s:vimwiki_get_known_syntaxes() " {{{
@@ -1361,6 +1363,15 @@ function! vimwiki#base#delete_link() "{{{
   " reread buffer => deleted wiki link should appear as non-existent
   if expand('%:p') != ""
     execute "e"
+  endif
+endfunction "}}}
+
+" vimwiki#base#one_column_one_row_toggle
+function! vimwiki#base#one_column_one_row_toggle() "{{{
+  if g:vimwiki_one_column_one_row
+      let g:vimwiki_one_column_one_row = 0
+  else
+      let g:vimwiki_one_column_one_row = 1
   endif
 endfunction "}}}
 
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -636,6 +636,7 @@ function! s:close_tag_table(table, ldest, header_ids) "{{{
       let n_cells = len(row)
       if n_cells > max_cells
         let max_cells = n_cells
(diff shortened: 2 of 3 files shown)


EinfachToll:
I agree that the restriction "one line in Vim == one table row" is a bit limiting. If I understand you correctly, with your solution, you can't make a table with multiple rows because everything is crammed in one cell, right? But a table with only one row is not really a table, is it? ;)
I think it would make more sense to introduce a row split marker, like there is currently a line to separate the table header from the body. For example, a table could look like this:

    | header       | header       |
    |==============|==============|
    | cell 1       | cell 2       |
    | still cell 1 |--------------|
    |--------------| cell 3       |
    | cell 4       | still cell 3 |

And a shortcut only to toggle a setting is really not necessary, IMO. It may be handy for some users, but probably too few to justify occupying one of the precious Vim keys.

standardeyre:
I tested my solution with multiple tables.  My intent was to ignore (not change) tables that have more than one column.  As near as I can tell, it works correctly. 

The "personal itch I was trying to scratch" was to hack at VimWiki in order to have some blocks of text "stand out" from the rest of the wiki text.  

My first thought of how to do that was to enclose the text in one column in a table.  For export from VimWiki to HTML (and then through pandoc to everything else), I did intend to create a table in HTML of one single row holding one single cell.  

I've since realized that I can get the effect I want by enclosing that block of text in a blockquote within VimWiki.  When I convert VimWiki to HTML I can modify the CSS for the blockquote to make that text further stand out.

I found this a great learning exercise to work on Vimscript.  I really like VimWiki.  The fork I have made _works for me_  :-) and I have it now.  If you want to tag this as something like "Feature Enhancement requested but denied" that's fine by me.

I look forward to contributing more to VimWiki in the future.

Thanks!

EinfachToll:
Ok. Maybe the option `g:vimwiki_valid_html_tags` is of interest for you, so you can directly insert HTML tags in the wiki file.

PR #337, opened 2017-04-21
Disable <leader>w(w,t,s,d,r) mappings with g:vimwiki_default_mappings

nexayq:
\<leader\>w is sometimes used by some other plugins or custom command.
This option allows user to use this plugin without worrying if his custom shortcut will be overridden.

You can disable \<leader\>w(w,t,s,d,r) mappings with setting:

    let g:vimwiki_default_mappings='0'

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,11 @@ normal mode:
  * `<Tab>` -- Find next wiki link
  * `<Shift-Tab>` -- Find previous wiki link
 
+You can disable <leader>w(w,t,s,d,r) mappings with setting:
+
+    let g:vimwiki_default_mappings='0'
+
+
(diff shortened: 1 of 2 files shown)


mmihira:
+1, would like this feature

pyrho:
https://github.com/vimwiki/vimwiki/pull/247 Seems to have a better approach.

EinfachToll:
What's wrong with the already existing option g:vimwiki_map_prefix?

johnmarcampbell:
It looks like the conversation here has died out, and as @EinfachToll there is already a way to do this with existing functionality. I'm closing this PR without merging. If anyone wants this PR to be merged we can reopen and discuss it.

PR #346, opened 2017-05-11
Support for data scheme

trnila:
Support for data URL scheme, so that `{{data:image/png;base64,...}}` could be used.


EinfachToll:
Thanks!

PR #347, opened 2017-05-14
Fix cursor positioning in vimwiki#lst#change_level

malramsay64:
In the function vimwiki#lst#change_level the cur_col was using the
column the cursor was in (`col('.')`). This is a problem when moving
from insert mode to normal mode however when the last column no longer
exists, this means that changing the level from the last column should
change the cursor position. This result is apparent in Neovim, but
doesn't occur in Vim.

By calculating cur_col from the last insert position this issue is
avoided in both vim and neovim.

### DIFF
--- a/autoload/vimwiki/lst.vim
+++ b/autoload/vimwiki/lst.vim
@@ -1057,7 +1057,7 @@ function! s:change_level(from_line, to_line, direction, plus_children) "{{{
 endfunction "}}}
 
 function! vimwiki#lst#change_level(from_line, to_line, direction, plus_children) "{{{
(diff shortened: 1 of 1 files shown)


EinfachToll:
Ok, but using '^ works only properly when you are movin lines in insert mode with Ctrl+d and Ctrl+t, but not with gll and glh in normal mode.

malramsay64:
Yeah that is somewhat problematic. Looking further into this, there is an [issue](https://github.com/neovim/neovim/issues/6962) on this problem in neovim, so I will leave it to be fixed there.

PR #358, opened 2017-06-07
If current wiki'syntax is 'markdown', there is no necessary to indent diary list by repeat(' ', &sw).

icymind:
The [markdown syntax rules](http://daringfireball.net/projects/markdown/syntax#list)  state:

List items may consist of multiple paragraphs. Each subsequent paragraph in a list item must be indented by either 4 spaces or one tab.

Diary list is top level at diary index, there is no necessary to indent diary list by any space. What's worse, if shiftwide is 4(that's normal for a markdown filetype), the diary list will indent 4 space, then the list will become a codeblock.

### DIFF
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -124,6 +124,14 @@ function! s:sort(lst) "{{{
   endif
 endfunction "}}}
 
+function! s:get_diary_indent() "{{{
+  if VimwikiGet('syntax') == 'markdown'
+    return ''
+  else
+    return repeat(' ', &sw)
+  endif
+endfunction "}}}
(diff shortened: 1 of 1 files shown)


EinfachToll:
Your reasoning is right, but now I remember that Vimwiki has already a per-wiki option `list_margin` and it would make sense to use the value of this option instead of `&sw`. Then, the user can set the value to 0 if the syntax is markdown. (It would make even more sense to make 0 the default of this option for markdown and media syntax.)

If you're inclined to change your patch accordingly, I'll happily merge it.

beuerle:
I implemented the `list_margin` approach and created a pull request #452.

Issue #359, opened 2017-06-18
Metadata placeholders


ycpei:
Hello,

Like `%title`, I thought perhaps a `%date` placeholder is useful too:
- html parser: if a line matches `^\s*%date\s.*$` then record what follows `^%date\s` as the date, and in the html template, replace `%date%` by the date.
- syntax highlight: add highlight for any line matching date placeholder.

This is useful at least for me because I often wonder when I wrote / edited a wiki file.

Also maybe `%authors` would be good too in case vimwiki is used for collaborative works... 

These changes may potentially also make vimwiki fit better in the Pandoc world because their AST has these metadata fields (of course a Pandoc vimwiki reader / writer is required).

PR #360, opened 2017-06-18
Added date placeholder.

ycpei:
See #359.

Let me know if there are any problems.


autoload/vimwiki/html.vim
```diff
@@ -1483,11 +1505,13 @@ function! s:convert_file(path_html, wikifile) "{{{
     call extend(ldest, lines)
 
     let title = s:process_title(placeholders, fnamemodify(a:wikifile, ":t:r"))
+    let date = s:process_date(placeholders, "")
```
EinfachToll: Today's date as a default value would make a lot of sense.

ycpei: Good point. I have made the change :)

EinfachToll:
Many thanks!

ycpei:
Thanks for merging!

PR #366, opened 2017-07-05
Predictable behaviour with multiple wikis

malramsay64:
This fixes all the irregularities in the movement between wikis and
diaries. Movement within the same wiki/diary can be done using the
`<Leader>ww`, `<Leader>wi`, `<Leader>w<Leader>w` commands without a number
prefix. The long form commands `:VimwikiDiaryIndex` etc. also work this
way.

To change between wikis/diaries either the shortened mappings or the
long form commands can be used.

This behaviour is achieved by passing a 0 by default to these commands,
which sets the index to that of the variable `g:vimwiki_current_idx`.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1310,8 +1310,7 @@ function! vimwiki#base#go_back_link() "{{{
   endif
 endfunction " }}}
 
-" vimwiki#base#goto_index
-function! vimwiki#base#goto_index(wnum, ...) "{{{
+function! vimwiki#base#get_index(wnum)
   if a:wnum > len(g:vimwiki_list)
     echomsg 'Vimwiki Error: Wiki '.a:wnum.' is not registered in g:vimwiki_list!'
     return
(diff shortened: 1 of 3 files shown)


EinfachToll:
Um, what are the irregularities you talked about?

malramsay64:
Sorry for being so vague in the first description, I didn't really have a well defined problem until going back to think about it now.

There are two different problems I am trying to fix;

1. The commands `:VimwikiDiaryIndex`, `:VimwikiIndex` etc. ignore any counts passed to them. This pull request provides a solution for this problem by using the `v:count` in the definition of the `<Plug>Vimwiki*` remaps as the count for the command they are calling.

2. When in a diary file, editing a diary file in another wiki takes running the command twice. If I am in the first wiki, `2<leader>w<leader>w` will change the variable `g:vimwiki_current_idx` to 1, however the buffer will not change until the second invocation of the key combination. This commit does nothing to solve this since I didn't have a good definition of the problem to start with and managed to merge the two problems.

I hope that makes things clearer

qubidt:
Is there any reason this PR was closed? It seems the irregular behavior still occurs.

malramsay64:
Because the 'fix' wasn't really a fix. See https://github.com/vimwiki/vimwiki/tree/refactor-internal-state-handling for current state of improved behavior.

PR #369, opened 2017-07-09
Possibility to increase/decrease checkbox

Nudin:
[Reopening of  #313…]

This implements Issue #308:
With :VimwikiToggleListItem we can change the status of a listitem between 0% and 100% – but we also include sub states configured in g:vimwiki_listsyms – for now we use them mostly to indicate the status of a item with subitems but it can already be nested, so it is possible to use the inbetween-states also for items without subitems.
This would become easier and more obvious if there are functions and keybindings to increment and decrement the state.

Adds new functions: `vimwiki#lst#increment_cb` & `vimwiki#lst#decrement_cb`
Adds keybinding for them: `gl+` &  `gl-`

### DIFF
--- a/ftplugin/vimwiki.vim
+++ b/ftplugin/vimwiki.vim
@@ -295,6 +295,8 @@ command! -buffer -range -nargs=1 VimwikiChangeSymbolTo call vimwiki#lst#change_m
 command! -buffer -range -nargs=1 VimwikiListChangeSymbolI call vimwiki#lst#change_marker(<line1>, <line2>, <f-args>, 'i')
 command! -buffer -nargs=1 VimwikiChangeSymbolInListTo call vimwiki#lst#change_marker_in_list(<f-args>)
 command! -buffer -range VimwikiToggleListItem call vimwiki#lst#toggle_cb(<line1>, <line2>)
+command! -buffer -range VimwikiIncrementListItem call vimwiki#lst#increment_cb(<line1>, <line2>)
+command! -buffer -range VimwikiDecrementListItem call vimwiki#lst#decrement_cb(<line1>, <line2>)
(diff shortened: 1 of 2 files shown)


ftplugin/vimwiki.vim
```diff
@@ -448,12 +448,12 @@ vnoremap <silent><script><buffer>
...
 if !hasmapto('<Plug>VimwikiIncrementListItem')
-  nmap <silent><buffer> <M-+> <Plug>VimwikiIncrementListItem
-  vmap <silent><buffer> <M-+> <Plug>VimwikiIncrementListItem
+  nmap <silent><buffer> gl+ <Plug>VimwikiIncrementListItem
+  vmap <silent><buffer> gl+ <Plug>VimwikiIncrementListItem
 endif
 if !hasmapto('<Plug>VimwikiDecrementListItem')
-  nmap <silent><buffer> <M--> <Plug>VimwikiDecrementListItem
-  vmap <silent><buffer> <M--> <Plug>VimwikiDecrementListItem
+  nmap <silent><buffer> gl- <Plug>VimwikiDecrementListItem
```
Nudin: There is actually a problem with gl-, since it is already bound to `:VimwikiChangeSymbolTo -`…

EinfachToll: Whoops, I didn't think of that. How about glj and glk, then? Although the keys glh and gll already do something unrelated, namely indenting and deindenting lines. And maybe someday we want to use glj and glk to move list items up and down. Or gln and glp? Although "next" and "previous" does not really make sense here. Still, I vote for the second option but I have no strong feelings about it.

Nudin: Not sure if moving lines up/down is something relevant to build in? There's the possibility to move a line with dd and p and with :m – but I also have not strong feelings. So gln and glp?

EinfachToll:
🍰

PR #377, opened 2017-08-17
Add horizontal alignment to table columns

schmeisers:
In the separator, you can now use `|------:|` to right align or `|:-----:|` to center align a column.


autoload/vimwiki/tbl.vim
```diff
@@ -234,6 +234,33 @@ function! s:get_rows(lnum) "{{{
...
+      endfor
+    else
+      let cells = vimwiki#tbl#get_cells(row)
+      for idx in range(len(cells))
+        if !has_key(aligns, idx)
+          let aligns[idx] = 'left'
+        endif
+      endfor
+    endif
+  endfor
```
EinfachToll: The alignments are set for every single row. It would be faster to break out of the loop if a separator was found and to set the default alignment only if no separator was found.

doc/vimwiki.txt
```diff
@@ -1658,6 +1658,15 @@ values: >
 
 To indent table indent the first row. Then format it with 'gqq'.
 
+You can specify horizontal alignment for columns in the separator.  The
```
EinfachToll: Better: "You can specify the type of horizontal alignment for columns in the separator using the ':' character."
(Just in case someone misses the `:`)

EinfachToll:
Nice work. If you feel like it, fix the two issues from my review.

hq6:
@schmeisers , I just tested this against the current `dev` branch, and it does not do right alignment on the first or last column on the following table. Can you please update this PR?

joejobber:
I just cloned the dev branch a few minutes ago to test this feature.

I found that in a table where the first column(s) is/are right-aligned that using <CR> in the last column of the last row, or <TAB> in the first column of the new row will left-align the right-aligned cells in the previous row. Further tabbing through the columns will correct the alignment. But if you go back to the first column <TAB> will left-align that column again.

It does this only in the second to last row after appending a new row. I did not yet test this with center-aligned columns.

joejobber:
Ouch, the key indicators were removed. I'll do it again and replace the angle brackets with right brackets. That'll teach me to post without previewing...

I found that in a table where the first column(s) is/are right-aligned that using [CR] in the last column of the last row, or [TAB} in the first column of the new row will left-align the right-aligned cells in the previous row. Further tabbing through the columns will correct the alignment. But if you go back to the first column [TAB] will left-align that column again.

It does this only in the second to last row after appending a new row. I did not yet test this with center-aligned columns.

ranebrown:
@joejobber I wasn't able to reproduce the described behavior but I might not be fully understanding the steps to reproduce. I think it might be better to open a new issue for further discussion. When you create the issue could you include an example table where you see the behavior? Can you also try using a minimal vimrc as described [here](https://vimwiki.github.io/vimwikiwiki/Troubleshooting.html#Troubleshooting-Minimal%20Vimrc)?

joejobber:
On 17/06/2019 21.36, Rane Brown wrote:
> @joejobber <https://github.com/joejobber> I wasn't able to reproduce the
> described behavior but I might not be fully understanding the steps to
> reproduce. I think it might be better to open a new issue for further
> discussion. 

OK. Will do.

> When you create the issue could you include an example table where
> you see the behavior? Can you also try using a minimal vimrc as described here
> <https://vimwiki.github.io/vimwikiwiki/Troubleshooting.html#Troubleshooting-Minimal%20Vimrc>?

No problem, but please give me some time. This week I have a deadline for the
next chapter of my new book to meet. Sometime the coming weekend, or at the
start of next week I'll try to provide a working example and a step-by-step
guide to reproduce.

A simple vimrc would be wise indeed. I wouldn't want to burden you with the one
I use for everyday editing. It's about a 1000 lines...

P.S. The code I sent as a diff in the original issue didn't show this behaviour,
but I haven't yet tried to find out what the difference is.

Grx HdV

joejobber:
I had some time in between things to do some tests. Sorry, this was with my normal vimrc, as this was an impromptu test session. Will try later again with a clean one when I find time to do a real test session. Thought maybe these results would help in the meantime.

HTH

[vimwiki_table_test.wiki.txt](https://github.com/vimwiki/vimwiki/files/3307385/vimwiki_table_test.wiki.txt)

ranebrown:
@joejobber sorry for the slow response. I think the behavior you are seeing is related to #47 and #658 which improved table formatting for large tables by only checking the row above the current row to format the table. Doing `gqq` should format the entire table as expected. Maybe we should add an option to disable that feature or at least add a note to the documentation.

joejobber:
I still hadn't found any time to look into this a bit deeper myself, so no apologies needed. However, I think you might be right in your assessment. I'll try the reformatting command. I think I did that before already and found it works, if I move the cursor a few rows up. But I am not sure and cannot verify at the moment. I'll try again when I am back and report on the results.

Thanks for looking into this. It is no biggie, but it would be nice if one wouldn't have to correct it manually.

joejobber:
Ok, I had some time for some more impromptu testing. I've documented the clean situation before the test and the result of the test in the file attached. I figure there are two things happening. The first is that I was used to having a separator line above and below a header, and one to close of the table. The opening and closing separator lines don't have alignment marks in them.  In the patch I submitted that worked as expected, and without thinking I presumed it would in the official code as well. Oh, well, I can get used to that. I'll remove those lines from all my tables. So, realizing this I also thought to test what happens if you try to have two alignment blocks in one table. Indeed, the second block is ignored. 

The other thing is that I think you are right and that the fixes for large tables are indeed "messing" with the alignment. I hope the test scenario's attached help (I had to rename it to txt for the request tracker, so you need to remove the extension or set the type manually if you want to use the test in vimwiki directly.).
[vimwiki_table_alignment_tests.wiki.txt](https://github.com/vimwiki/vimwiki/files/3389737/vimwiki_table_alignment_tests.wiki.txt)

PR #380, opened 2017-08-18
Escape the 'dot' in extension substitution

Dru89:
This escapes the `.` preceding an extension (e.g. `.wiki`) so that it doesn't accidentally match any character.

For example: `[vimwiki](vimwiki)` was opening the `vi.wiki` page, because of this rule.


Dru89:
This bug also exists in `master`, but I'm not sure what the strategy is for updating bugs in the non-dev branch.

https://github.com/vimwiki/vimwiki/blob/master/autoload/vimwiki/markdown_base.vim#L112

Nudin:
Looks good. Thanks for contributing!
A Patch for master is not needed since that branch is not developed further anymore (for now). But we'll need to port it to "refactor-internal-state-handling" once it's done. But this patch is so tiny it's no problem. ;)

PR #390, opened 2017-09-16
Allow to specify additional chars for lists

Nudin:
I'd like to use additional characters for the bullets of lists especially unicode-chars like '→':
→ foo
→ bar
→ blub

This patch allows to set a variable g:vimwiki_additional_bullet_types to do so.
In HTML-export they appear as normal bullets and it does create keybindings like gl<char> for those since this would be nosense for chars not on the keyboard.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1455,7 +1455,11 @@ function! s:convert_file(path_html, wikifile) "{{{
     endif
 
     " prepare regexps for lists
-    let s:bullets = '[*-]'
+    if exists("g:vimwiki_additional_bullet_types")
+      let s:bullets = '[*-'. join(keys(g:vimwiki_additional_bullet_types), '') . ']'
+    else
+      let s:bullets = '[*-]'
+    endif
     let s:numbers =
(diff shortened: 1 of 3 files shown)


sharethewisdom:
I'd like to use `dev`'s HEAD with this feature included. But I checked out this commit last week and I think I remember it being reverted because it 'd break stuff. Is that correct?

Nudin:
> But I checked out this commit last week and I think I remember it being reverted because it 'd break stuff. Is that correct?

Yes and no. This patch didn't break anything but while merging it in something went wrong. Thanks for the reminder. Will add it again.

Nudin:
@sharethewisdom @oguzserbetci See  #479.

EinfachToll:
Ich hab deine letzten Commits mal in einen eigenen Branch (configurable_list_syms) verschoben, weil es Fehler gab (#602). Keine Katastrophe, aber nächstes Mal bitte ein kleines bisschen Aufwand investieren, deine Änderungen zu testen.
Übrigens, deine Methode, listsyms und listsyms_rejected sowohl global als auch wikilokal einstellbar zu machen ist unnötig, das ging vorher auch schon, wenn du dir s:populate_wikilocal_options() mal anschaust. Ein schönes, undokumentiertes Feature ;)

PR #393, opened 2017-09-17
Add License-File and update contributers-list

Nudin:
Vimwiki currently has not License-File, and the license is only mentioned in the help-file. We should change this.

Also it would be ideal to add a license header on top of every source-code-file.

Still open is the question: who should be named as primary copyright-holder and for which years?
Probably something like:

```
2008-2013 Maxim Kim
2013-2017 EinfachToll/Your real name
```

See also #386.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) 2008-2010 Maxim Kim
+              2013-2017 Daniel Schemala
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2008-2010 Maxim Kim
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


johnmarcampbell:
This looks good to me.
1) @EinfachToll : Do you want your real name put in the license file? Or leave it as your github handle?  What are the right years?
2) Minor point: The help file has the phrase "The MIT Licence" in it. I know "licence" is a legitimate European spelling, but since here the word "License" is part of the title perhaps the help file should be changed?

EinfachToll:
> Also it would be ideal to add a license header on top of every source-code-file.

Is this really necessary? AFAIK, it suffices to mention the licensing stuff _somewhere_ in the project, just so that somebody using the project in an illegal way has a hard time explaing they didn't know it was licensed. Apart from the minor fact that this is ... rather unlikely to happen in the case of Vimwiki, a file with the name `LICENSE` sitting in the top directory should suffice.

> Still open is the question: who should be named as primary copyright-holder and for which years?

Your suggestion looks good to me. You can look up my real name in the contributors section in the help file.

> but since here the word "License" is part of the title perhaps the help file should be changed?

Yes, "Licence" was probably only a typo.

Nudin:
>> Also it would be ideal to add a license header on top of every source-code-file.
>
> Is this really necessary? AFAIK, it suffices to mention the licensing stuff somewhere in the project, just so that somebody using the project in an illegal way has a hard time explaing they didn't know it was licensed. Apart from the minor fact that this is ... rather unlikely to happen in the case of Vimwiki, a file with the name LICENSE sitting in the top directory should suffice.

It's not really necessary, but strongly recommended. The main reason is, that the longer an open source project exists, the more likely it is that it is forked, combined with coder from other projects, parts of the code is reused in other projects, etc. This often leads to confusion, which license applies to what, etc.

>>  but since here the word "License" is part of the title perhaps the help file should be changed?
>
> Yes, "Licence" was probably only a typo.

It's American English vs British English, so no big deal, but I agree that "Licence" is the common way and the way OSI spells it. Fixed it.

johnmarcampbell:
I agree that the license file alone is sufficient. I can see the argument for putting it in every source file, but from a practical perspective I think it's unnecessary. And from an aesthetic perspective I think it looks nicer just to have the license info in one place at the root of the project.

All that said, I think this PR looks good and can be merged in.

Issue #394, opened 2017-09-17
Link creating with single quote fails


breneser:
Hi, when I press enter on a chapter title with single quote (') to create a link I get following error:

> 
> Error detected while processing function vimwiki#base#normalize_link[3]..vimwiki#markdown_base#normalize_link[8]..<SNR>197_normalize_link_syntax_v:
> line   11:
> E15: Invalid expression: 'Sprucing Up Your Site's Navigation'
> Error detected while processing function vimwiki#base#normalize_link[3]..vimwiki#markdown_base#normalize_link:
> line    8:
> E171: Missing :endif
> Error detected while processing function vimwiki#base#normalize_link:
> line    3:
> E171: Missing :endif
>

johnmarcampbell:
@breneser I wasn't able to reproduce this problem. Furthermore, the phrase 'Sprucing Up Your Site's Navigation' doesn't appear anywhere in vimwiki's source code.
1) Are you using the latest version?
2) If so, could you provide a concrete example that triggers this error?

breneser:
@johnmarcampbell It has been a while I will check if I have the latest version and if this is still the case.

That phrase is not in the source code, just some notes I have been taking with Vimwiki :)

johnmarcampbell:
Ahh, I see. I thought that phrase was somehow part of the internal error message. I see now that it's probably the text you were trying to make a link out of. Thanks for clarifying.

EinfachToll:
I can reproduce this error: just type the text `foo'bar` into a Vimwiki file, select it in visual mode and press enter. The problem is that in autoload/vimwiki/base.vim in the function s:normalize_link_syntax_v(), the `@"` (which is a special variable in Vimscript which holds the content of the `"` register) needs to be escaped properly. Currently, if it contains a `'`, the concatenation blows up.
(^- Hint for my fellow developers ;)

johnmarcampbell:
Okay, that makes sense. I'm able to reproduce the problem now. (I'm honestly not sure why I couldn't before :man_shrugging:). I'll take a look at it and see if I can't fix it.

PR #405, opened 2017-10-26
Fix link generation around apostrophes; issue #394

johnmarcampbell:
I modified the `s:normalize_link_syntax_v()` function in `autoload/vimwiki/base.vim`. This function contains a call to `substitute()` with an argument of `'\=' . "'" . @" . "'"`, which interprets the contents of `@"` as a an expression. This is normally fine, but when `@"` has an apostrophe in it (e.g. *foo'bar*), it doesn't evaluate to a valid expression (e.g. *'foo'bar'*).

So I removed the `'\='`, etc., and just have `substitute` operate on the contents of the register directly. This works for all the cases I tested, and doesn't break anything as far as I can tell. If there's a good reason the `'\='`, etc., should stay in, let me know and I'll find a different way.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1996,7 +1996,7 @@ function! s:normalize_link_syntax_v() " {{{
       let sub = s:normalize_link_in_diary(@")
     else
       let sub = substitute(g:vimwiki_WikiLinkTemplate1,
-            \ '__LinkUrl__', '\=' . "'" . @" . "'", '')
+            \ '__LinkUrl__', @", '')
     endif
 
     " Put substitution in register " and change text


johnmarcampbell:
@Nudin or @EinfachToll, would one of you be able to vet this PR? I just don't want to merge it without at least one more pair of eyes checking it out.

Nudin:
This does not work ether, since now things like `foo&bar` or `foo\bar` (subfolders on windows) will fail. See `:help sub-replace-special`.

The only solution I found so far is:
```
substitute(g:vimwiki_WikiLinkTemplate1, '__LinkUrl__',
                '\=' . "'" . substitute(@", "'", "''", 'g') . "'", '')
```

Nudin:
#409 is basically the same.

Nudin:
There are actually a lot of these in the code:

```
autoload/vimwiki/markdown_base.vim
142:    let link = substitute(g:vimwiki_Weblink1Template, '__LinkUrl__', '\='."'".visual_selection."'", '')
143:    let link = substitute(link, '__LinkDescription__', '\='."'".visual_selection."'", '')

autoload/vimwiki/base.vim
468:            \ substitute(g:vimwiki_WikiLinkTemplate1, '__LinkUrl__', '\='."'".link."'", ''))
683:          \ '__Header__', "\\='".segment."'", '')
685:          \ '__Text__', "\\='".segment."'", '')
687:          \ '__Tag__', "\\='".segment."'", '')
1181:        \ '__Header__', '\='."'".a:start_header."'", '')
1831:          \ '\='."'".'#'.esc_link."'", '')
1832:    let link = substitute(link, '__LinkDescription__', '\='."'".esc_desc."'", '')
1853:    let lnk = substitute(lnk, '__LinkUrl__', '\='."'".a:rxUrl."'", 'g')
1856:    let lnk = substitute(lnk, '__LinkDescription__', '\='."'".a:rxDesc."'", 'g')
1859:    let lnk = substitute(lnk, '__LinkStyle__', '\='."'".a:rxStyle."'", 'g')
1898:  let lnk = substitute(template, '__LinkDescription__', '\="'.descr.'"', '')
1899:  let lnk = substitute(lnk, '__LinkUrl__', '\="'.url.'"', '')
1907:  let lnk = substitute(lnk, '__LinkStyle__', '\="'.style.'"', '')
2000:            \ '__LinkUrl__', '\=' . "'" . @" . "'", '')
2031:        \ 'substitute(v:val, last_word, "\\=submatch(1)", "")')
```
We should probably fix them all by adding a function doing the escaping.

Nudin:
See #411 for a more general fix of the problems.

PR #408, opened 2017-11-03
Change directory when switching to another wiki

BirgerNi:
I have multiple wikis. When I switch to another wiki via `<Leader>ws` or `<Plug>VimwikiUISelect` the working directory does not change. This is problematic because I often use `:fin` to jump to another wiki file.

This patch changes the working directory to the directory where is wiki is in.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1337,6 +1337,7 @@ function! vimwiki#base#goto_index(wnum, ...) "{{{
         \ VimwikiGet('ext', idx),
         \ '')
   call vimwiki#base#setup_buffer_state(idx)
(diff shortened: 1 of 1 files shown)


johnmarcampbell:
I don't know if this is desirable default behavior. I imagine many users (myself included) might not want the working directory to change every time a new wiki is selected. Could this be solved with a mapping in your .vimrc along the lines of `autocmd BufRead index.wiki cd %:h`? Or just `nnoremap <Leader>cd :cd %:h<CR>`, which gives you a quick way to switch working directories even outside of vimwiki.

BirgerNi:
`autocmd BufRead index.md cd %:h` works like a charm. Thank you!

Issue #338, opened 2017-04-22
Apostrophe in anchor caused error


Flurrywinde:
I had a header with an apostrophe in it, something like:
`== A dog's life ==`
Then I made a link to it: `[[Some file#A dog's life]]`

Got this error: 
> Error detected while processing function vimwiki#base#follow_link[39]..vimwiki#base#open_link[26]..vimwiki#base#edit_file[34]..<SNR>74_jump_to_anchor:
>line 9:
> E15: Invalid expression: 'A dog's life'

johnmarcampbell:
I just tested this with PR #411 merged in, and it works fine. @Flurrywinde, if you're experiencing this problem can you update *from the dev branch* and see if it fixes it?

Flurrywinde:
Works fine now with the latest dev branch. Thanks!

johnmarcampbell:
Glad to hear it!

PR #409, opened 2017-11-04
Fix bug #338

Nudin:
Fixing:
#338: Apostrophe in anchor caused error

This is similar to #405.


Nudin:
Closing since superseded by more general solution #411.

PR #411, opened 2017-11-10
Use a safe substitution-function everywhere

### LINKED ISSUE #394: Link creating with single quote fails

Nudin:
This fixes #394 #405 #338 #409 as well probably others and removes some up duplicate code.
It contains the new function safesubstitute twice since I didn't want to make it public.

This code should be reviewed and tested again. @johnmarcampbell Can you try it out and test it with some special-characters? I tested it quite a bit, but maybe I missed something…

### DIFF
--- a/autoload/vimwiki/markdown_base.vim
+++ b/autoload/vimwiki/markdown_base.vim
@@ -6,6 +6,13 @@
 
 " MISC helper functions {{{
 
+" s:safesubstitute
+function! s:safesubstitute(text, search, replace, mode) "{{{
+  " Substitute regexp but do not interpret replace
+  let escaped = escape(a:replace, '\&')
+  return substitute(a:text, a:search, escaped, a:mode)
+endfunction " }}}
+
 " vimwiki#markdown_base#reset_mkd_refs
 function! vimwiki#markdown_base#reset_mkd_refs() "{{{
   call VimwikiClear('markdown_refs')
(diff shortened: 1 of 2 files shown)


johnmarcampbell:
Just to comment: I tested it with a bunch of special characters, and it performed well with everything I threw at it.

autoload/vimwiki/markdown_base.vim
```diff
@@ -139,8 +146,8 @@ function! s:normalize_link_syntax_v() " {{{
   try
     norm! gvy
     let visual_selection = @"
-    let link = substitute(g:vimwiki_Weblink1Template, '__LinkUrl__', '\='."'".visual_selection."'", '')
-    let link = substitute(link, '__LinkDescription__', '\='."'".visual_selection."'", '')
+    let link = Safesubstitute(g:vimwiki_Weblink1Template, '__LinkUrl__', visual_selection, '')
```
dangkhoasdc: It should be s:safesubstitute (instead of Safesubstitute), right? 

I got an error when create a note with multiple words in the title. (unknown function SafeSubstitute, my workaround is to change to `s:safesubstitute`. However, I am not sure whether it is a solution.

johnallen3d: I'm seeing this error as well.

Nudin: Sorry for long off-time. You're absolutely right.

Nudin: Fixed by #418

PR #442, opened 2018-02-23
Fix path resolution of diary notes

blyoa:
## What does vimwiki do?

The vimwiki fogets `wnum` passed to `vimwki#diary#make_note` before opens a diary note.
The vimwiki opens a default diary note path, therefore, even if we configured `g:vimwiki_list` and passed `[count]` to `<Plug>VimwikiMakeDiaryNote`.

## What do I expect?

I expect that the vimwiki treats `[count]` and opens a diary note associated to the number of `[count]`.

## What is a problem?

The problem is that we cannot configure the paths of diary notes via `g:vimwiki_list`.

## What will this commit do?

This commit will fix the problem by allowing `vimwki#base#open_link` to use an already resolved diary path in `vimwiki#diary#make_note`.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -253,7 +253,12 @@ endfunction "}}}
 
 " vimwiki#base#open_link
 function! vimwiki#base#open_link(cmd, link, ...) "{{{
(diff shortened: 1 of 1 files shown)


blyoa:
I believe this pull request fixes also the issue #451.

PR #444, opened 2018-02-25
Multiline bullet-lists, HTML comments and TODO

silvioricardoc:
This patch changes a couple of little details. I split it in 3 commits, but maybe I should have split it in 3 patches? Anyway, here they are:

1. Allow HTML-style comment syntax, such as `<!-- my comment -->`, which is officially part of markdown.

2. Treat highlighted bullet-point entries spanning multiple lines as a single entity. This happens when we have a checked [X] item whose text continues on the next line.

3. Avoid matching "TODO" on word boundaries (e.g. does not match "xTODO:"). Also, with this patch, the ":" after TODO is not required (so e.g. "[TODO]" also works).

### DIFF
--- a/autoload/vimwiki/vars.vim
+++ b/autoload/vimwiki/vars.vim
@@ -164,7 +164,7 @@ function! s:populate_global_variables()
   let g:vimwiki_global_vars.rxWikiInclSuffix1 = g:vimwiki_global_vars.rxWikiInclArgs.
         \ g:vimwiki_global_vars.rxWikiInclSuffix
 
-  let g:vimwiki_global_vars.rxTodo = '\C\%(TODO:\|DONE:\|STARTED:\|FIXME:\|FIXED:\|XXX:\)'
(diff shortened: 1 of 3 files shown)


EinfachToll:
1. and 3. are fine, but I don't get number 2. Can you post an example when a bullet point is not threated correctly?

silvioricardoc:
```
* [X] This example should be highlighted
      as a single entity, meaning that all of these lines
      should be shown as a Comment when [X] is checked.
* [ ] This line should not be highlighted, OTOH.
```

EinfachToll:
There is already the option `g:vimwiki_hl_cb_checked`. Setting this to 2 should do what you want. Or is there still a difference?

silvioricardoc:
Oh wow, I completely missed that. It works great when I set it to 2, so I agree we can skip number (2) above.

Would you still accept (1) and (3)? Should I create another pull request with only these?

EinfachToll:
Yes, (1) and (3) are ok. You probably don't need to open a new PR, but you can perhaps add a new commit to your `multiline-and-comments` branch where you revert (2) or something like that. Github became a bit complicated over the last years 😮

EinfachToll:
👍

PR #457, opened 2018-03-17
Add PrismJS support

ruchee:
Let syntax highlight tag custom

### DIFF
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -143,11 +143,11 @@ function! s:format_diary() "{{{
         if empty(cap)
           let entry = substitute(vimwiki#vars#get_global('WikiLinkTemplate1'), '__LinkUrl__', fl, '')
           let entry = substitute(entry, '__LinkDescription__', cap, '')
-          call add(result, repeat(' ', &sw).'* '.entry)
+          call add(result, repeat(' ', vimwiki#lst#get_list_margin()).'* '.entry)
         else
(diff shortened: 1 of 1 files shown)


EinfachToll:
Looks good so far, but can you change the two new options so that instead of two strings, it is one option `g:vimwiki_code_highlight_tags` which is a list of two strings (for start and end tags, obviously)? And then add a note in the help file?

autoload/vimwiki/html.vim
```diff
@@ -786,22 +786,31 @@ function! s:process_tag_pre(line, pre) "{{{
   let lines = []
   let pre = a:pre
   let processed = 0
+
+  let code_tag_begin = '<pre>'
+  let code_tag_end   = '</pre>'
+  if exists('g:vimwiki_code_highlight_tag_begin') && exists('g:vimwiki_code_highlight_tag_end')
```
EinfachToll: Can you change the two new options so that instead of two strings, it is one option `g:vimwiki_code_highlight_tags` which is a list of two strings (for start and end tags, obviously)?

doc/vimwiki.txt
```diff
@@ -1193,6 +1193,23 @@ Result of HTML export: >
...
+         print("Hello {0} number {1}".format(world, x))
+}}}
+
+Result of HTML export: >
+ <pre><code class="language-python">
+ def hello(world):
+     for x in range(10):
+         print("Hello {0} number {1}".format(world, x))
+ </code></pre>
+
```
EinfachToll: Can you add a note about the new option in the doc?

PR #459, opened 2018-03-19
Customize the generated tag name for code block

ruchee:
Combine two options into one

Tks @EinfachToll

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -786,6 +786,14 @@ function! s:process_tag_pre(line, pre) "{{{
   let lines = []
   let pre = a:pre
   let processed = 0
+
+  let code_tag_start = '<pre>'
+  let code_tag_end   = '</pre>'
+  if exists('g:vimwiki_code_highlight_tags')
+    let code_tag_start = g:vimwiki_code_highlight_tags['start']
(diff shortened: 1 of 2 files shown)


autoload/vimwiki/html.vim
```diff
@@ -786,22 +786,31 @@ function! s:process_tag_pre(line, pre) "{{{
   let lines = []
   let pre = a:pre
   let processed = 0
+
+  let code_tag_start = '<pre>'
+  let code_tag_end   = '</pre>'
+  if exists('g:vimwiki_code_highlight_tags')
```
EinfachToll: No need to check if it exists. Initialize the variable with a default value instead. This is done in the file `autoload/vimwiki/vars.vim` in the function `s:populate_global_vars`. See how the other options are initialized.

autoload/vimwiki/html.vim
```diff
@@ -786,22 +786,31 @@ function! s:process_tag_pre(line, pre) "{{{
...
+    let code_tag_start = g:vimwiki_code_highlight_tags['start']
```
EinfachToll: Don't read the option variable directly, use the function `vimwiki#vars#get_global()` instead.

hq6:
I am closing this PR because it looks like there was no response to review, and it looks like the author has removed the fork. If there is renewed interest in this, please make a new PR.

PR #474, opened 2018-04-08
Fix bug #467 and improve handling of anchors

### LINKED ISSUE #467: Links should not be escaped in HTML-output
a link like `[[foo/bar|baz]]` is currently rendered to `<a href="foo&#47;bar.html">baz</a>` instead of `<a href="foo/bar.html">baz</a>`. The links still work in all tested browsers but are technically wrong.

Nudin:
Until now the function `safe_html_anchor` was applied to the whole link. Now it's only applied to the anchor. So things in the filename don't get encoded anymore. Fixes #467 

@EinfachToll  What do you think?

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -370,6 +370,7 @@ endfunction
 
 function! vimwiki#html#linkify_link(src, descr) "{{{
   let file_str = split(a:src, '#')[0]
+  let file_str = substitute(file_str, '"', '\&quot;', 'g')
   let anchor_str = join(split(a:src, '#')[1:], '%23')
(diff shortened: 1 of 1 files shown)


autoload/vimwiki/html.vim
```diff
@@ -370,6 +370,7 @@ endfunction
 
 function! vimwiki#html#linkify_link(src, descr) "{{{
   let file_str = split(a:src, '#')[0]
+  let file_str = substitute(file_str, '"', '\&quot;', 'g')
```
Nudin: To avoid breaking of HTML when the filename contains a `"`… (although you'll have other issues in that case anyway)

EinfachToll:
No need to merge this, because it's now solved in a different way.

PR #479, opened 2018-04-23
Allow to specify additional chars for lists

Nudin:
Reimplementation of #390 (41e653718e29933dd9559e546b859158cc87804a).

From the original:
I'd like to use additional characters for the bullets of lists especially unicode-chars like '→':
→ foo
→ bar
→ blub

This patch allows to set a variable g:vimwiki_additional_bullet_types to do so.
In HTML-export they appear as normal bullets and it does create keybindings like gl for those since this would be nosense for chars not on the keyboard.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1482,7 +1482,11 @@ function! s:convert_file(path_html, wikifile)
     endif
 
     " prepare regexps for lists
-    let s:bullets = '[*-]'
+    if exists("g:vimwiki_additional_bullet_types")
+      let s:bullets = '[*-'. join(keys(g:vimwiki_additional_bullet_types), '') . ']'
+    else
+      let s:bullets = '[*-]'
+    endif
     let s:numbers = '\C\%(#\|\d\+)\|\d\+\.\|[ivxlcdm]\+)\|[IVXLCDM]\+)\|\l\{1,2})\|\u\{1,2})\)'
 
     for line in lsource
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1483,7 +1483,7 @@ function! s:convert_file(path_html, wikifile)
 
     " prepare regexps for lists
     if exists("g:vimwiki_additional_bullet_types")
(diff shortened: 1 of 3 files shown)


Nudin:
This is still WIP.

autoload/vimwiki/vars.vim
```diff
@@ -340,12 +340,16 @@ function! vimwiki#vars#populate_syntax_vars(syntax)
         \ '^\s*'.g:vimwiki_syntax_variables[a:syntax].rxMathEnd.'\s*$'
 
   " list stuff
+  let l:bullet_types = g:vimwiki_syntax_variables[a:syntax].bullet_types
```
Nudin: @EinfachToll What do you think about using a local variable here for this? I'm not happy but didn't come up with a better way.

autoload/vimwiki/html.vim
```diff
@@ -1482,7 +1482,11 @@ function! s:convert_file(path_html, wikifile)
     endif
 
     " prepare regexps for lists
-    let s:bullets = '[*-]'
+    if exists("g:vimwiki_additional_bullet_types")
+      let s:bullets = '[*-'. join(g:vimwiki_additional_bullet_types, '') . ']'
```
EinfachToll: Don't forget to escape properly when you use an arbitrary string in a regex. According to the Vim doc (`:h /\]`), the characters `]^-\` must be escaped.

oguzserbetci:
Do you need any help? I would like to have this feature and am ready to help :) @Nudin

Nudin:
@oguzserbetci hey I didn't had any time to finish this. If you're interested in finishing this, that would be great! The first small thing is the second comment of EinfachToll above. That's an easy thing (ideal to start).
Then EinfachToll and I agreed to have not a single variable containing additional chars but rather have an option per wiki with which you can set the chars. This list should have the current values as default (and not as currently append it to the default-list), so that by changing it you can actually also disable characters, set by the default. For reading the options there are usable functions already available.

If you need help or if anything is unclear just ask!

Nudin:
I'm currently rewriting this.

PR #492, opened 2018-05-12
setting script variable for autowrite_all, was missing before

huynle:
Vim complains about `s:vimwiki_autowriteall_saved` not being set when `g:vimwiki_autowriteall` has been previous set during loading of plugin.

### DIFF
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -13,6 +13,7 @@ let g:loaded_vimwiki = 1
 let s:old_cpo = &cpo
 set cpo&vim
 
+let s:vimwiki_autowriteall_saved = g:vimwiki_autowriteall
 
 " this is called when the cursor leaves the buffer
 function! s:setup_buffer_leave()


EinfachToll:
Ok, but out of curiosity, what would you do to trigger an error here?

huynle:
The error gets triggered after I load a saved vim session, then `:Vimiwiki`, then move to another window.
```bash
E121: Undefined variable: s:vimwiki_autowriteall_saved
E15: Invalid expression: s:vimwiki_autowriteall_saved
Error detected while processing function vimwiki#base#follow_link[58]..vimwiki#base#open_link:
line   31:
E171: Missing :endif
Error detected while processing function vimwiki#base#follow_link:
line   58:
E171: Missing :endif
```

hiberabyss:
@huynle @EinfachToll There need protection for the assignment if there is no `g:vimwiki_autowriteall` defined.

Nudin:
Since there were several reports of broken functionality I reverted this change.

Issue #154, opened 2015-08-26
Diary index creation lists all headlines


btwe:
I was hacking on master right away to change the behavior of :VimwikiDiaryGenerateLinks
The patch will now list all headings, which I at least find useful.

Sorry, I won't create an github repo for that so I'll just paste the patch below. If you like you could just copy paste.

``` diff
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -11,8 +11,6 @@ endif
 let g:loaded_vimwiki_diary_auto = 1
 "}}}

-let s:vimwiki_max_scan_for_caption = 5
-
 " Helpers {{{
 function! s:prefix_zero(num) "{{{
   if a:num < 10
@@ -78,18 +76,19 @@ fun! s:read_captions(files) "{{{
   for fl in a:files
     " remove paths and extensions
     let fl_key = fnamemodify(fl, ':t:r')
+    let fl_captions = []

     if filereadable(fl)
-      for line in readfile(fl, '', s:vimwiki_max_scan_for_caption)
-        if line =~ g:vimwiki_rxHeader && !has_key(result, fl_key)
-          let result[fl_key] = vimwiki#u#trim(matchstr(line, g:vimwiki_rxHeader))
+      "for line in readfile(fl, '', s:vimwiki_max_scan_for_caption)
+      for line in readfile(fl, '')
+        if line =~ g:vimwiki_rxHeader
+          let curcap = vimwiki#u#trim(matchstr(line, g:vimwiki_rxHeader))
+          call add(fl_captions, curcap)
         endif
       endfor
     endif

-    if !has_key(result, fl_key)
-      let result[fl_key] = ''
-    endif
+    let result[fl_key] = fl_captions

   endfor
   return result
@@ -163,15 +162,17 @@ fun! s:format_diary(...) "{{{
       call add(result, substitute(g:vimwiki_rxH3_Template, '__Header__', s:get_month_name(month), ''))

       " for [fl, cap] in s:rev(sort(items(g_files[year][month])))
-      for [fl, cap] in s:sort(items(g_files[year][month]))
-        if empty(cap)
+      for [fl, caplist] in s:sort(items(g_files[year][month]))
+        if empty(caplist)
           let entry = substitute(g:vimwiki_WikiLinkTemplate1, '__LinkUrl__', fl, '')
           let entry = substitute(entry, '__LinkDescription__', cap, '')
[...]

PR #498, opened 2018-05-23
Added optional sub-captions to diary index.

edlanglois:
Added the option 'diary_caption_level'.
At -1 (the default) the diary index is the same as before.

At level 1, the diary page links have no caption but have a sub-list of
captions linking to the individual top-level headers within the diary
page.

Example:
```
=== April ===
  * [[2018-04-30]]
    - [[2018-04-30#My Section|My Section]]
  * [[2018-04-27]]
    - [[2018-04-27#Some Section|Some Section]]
    - [[2018-04-27#Another Section|Another Section]]
```

At level 2, the main caption is the first top-level header and
sub-captions link to 2nd-level headers.

Example:
```
=== April ===
  * [[2018-04-30|My Section]]
    - [[2018-04-30#My Subsection|My Subsection]]
  * [[2018-04-27|Some Section]]
    - [[2018-04-27#Part 1|Part 1]]
    - [[2018-04-27#Part 2|Part 2]]
```

Inspired by: https://github.com/vimwiki/vimwiki/issues/154#issuecomment-159900916
Also related: https://github.com/vimwiki/vimwiki/pull/30

### DIFF
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -63,26 +63,98 @@ function! s:get_month_name(month)
   return vimwiki#vars#get_global('diary_months')[str2nr(a:month)]
 endfunction
 
+function! s:get_first_header(fl)
+  " Get the first header in the file within the first s:vimwiki_max_scan_for_caption lines.
+  let header_rx = vimwiki#vars#get_syntaxlocal('rxHeader')
+
+  for line in readfile(a:fl, '', s:vimwiki_max_scan_for_caption)
+    if line =~# header_rx
+      return vimwiki#u#trim(matchstr(line, header_rx))
+    endif
+  endfor
+  return ''
+endfunction
+
+function! s:get_all_headers(fl, maxlevel)
+  " Get a list of all headers in a file up to a given level.
+  " Returns a list whose elements are pairs [level, title]
+  let headers_rx = {}
+  for i in range(1, a:maxlevel)
+    let headers_rx[i] = vimwiki#vars#get_syntaxlocal('rxH'.i.'_Text')
+  endfor
+
+  let headers = []
+  for line in readfile(a:fl, '')
+    for [i, header_rx] in items(headers_rx)
+      if line =~# header_rx
+        call add(headers, [i, vimwiki#u#trim(matchstr(line, header_rx))])
+        break
+      endif
+    endfor
+  endfor
+  return headers
+endfunction
+
+function! s:count_headers_level_less_equal(headers, maxlevel)
+  " Count headers with level <=  maxlevel in a list of [level, title] pairs.
+  let l:count = 0
+  for [header_level, _] in a:headers
+    if header_level <= a:maxlevel
+      let l:count += 1
+    endif
+  endfor
+  return l:count
+endfunction
+
+function! s:get_min_header_level(headers)
+  " The minimum level of any header in a list of [level, title] pairs.
+  if len(a:headers) == 0
+    return 0
+  endif
+  let minlevel = a:headers[0][0]
+  for [level, _] in a:headers
+    let minlevel = min([minlevel, level])
+  endfor
+  return minlevel
+endfunction
+
 
 function! s:read_captions(files)
   let result = {}
-  let rx_header = vimwiki#vars#get_syntaxlocal('rxHeader')
+  let caption_level = vimwiki#vars#get_wikilocal('diary_caption_level')
+
   for fl in a:files
     " remove paths and extensions
-    let fl_key = substitute(fnamemodify(fl, ':t'), vimwiki#vars#get_wikilocal('ext').'$', '', '')
+    let fl_captions = {}
 
-    if filereadable(fl)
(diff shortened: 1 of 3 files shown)


EinfachToll:
The suggested behavior looks a bit muddled to me. When I set this option to 2 but then I have multiple level-1 headers in the diary page, only the first one is taken for the caption and each level-2 header is added as subitem in the diary index, even the ones which are under other level-1 headers.
How about the following behavior for the various values of the new option:
- `-1`: no special captions and list items at all
- `0`: the behavior as it is now, that is, the first header is taken as caption
- `1`: if there is exactly one level-1 header, it is taken as the caption, else, they are added as subitems.
- `2`: if there is exactly one level-1 header, it is taken as the caption and the level-2 headers are added as subitems. If there are more level-1 headers, they are added as subitems and the level-2 headers as sub-sub-items
- `3` and so on up to level-3 headers
- `4` etc.

edlanglois:
Sounds good, I was aiming for consistency in the caption level as you vary the number of headers in the diary page but I agree that the current behaviour with multiple level-1 headers when set to 2 isn't great. I'm happy with either approach so I'll update it to work as you suggest.

edlanglois:
I've updated the pull request to match the behaviour you requested. The commit message has a description of the behaviour:

> Added the option 'diary_caption_level', which controls the presence of
> captions in the diary index linking to headers within the diary pages.
> 
> Possible values:
> -1:  No headers are read from the diary page.
>  0:  The first header from the diary page is used as the caption.
>        There are no sub-captions.
>  1:  Captions are created for headers of level 1 in the diary page.
>  2:  Captions are created for headers up to level 2 in the diary page.
>  etc.
> 
> When the value is >= 1, the primary caption of each diary page is set to
> first header read from that page if it is the unique lowest-level
> header.

I didn't document this but there was also the question of what to do about skipped header levels. The caption indentation depth is shifted over so that the smallest indentation depth is one deeper than the main page link. So you won't end up with all of the sub-captions indented 3 deep if all headers in the diary page are at 3rd level instead of 1st. Other than that shift, indentation depth is preserved relative to header level.

hq6:
I am a new maintainer and I'm interested in getting this merged.

@edlanglois , if you are still interested in this PR, please resolve the conflicts and test against the tip of `dev`.

edlanglois:
Rebased on dev and tested it, everything still works for me.

I cleaned up the change a bit as well:
* fixed endfo => endfor (somehow that worked in the first place?)
* removed redundant "l:" variable scopes
* removed an unused optional argument to get_first_header
* added missing "the" to the documentation

hq6:
I am not super familiar with the diary feature, so it's possible that I am misusing this, but I was not able to reproduce the intended behavior of this patch. Here's what I tried.

1. Put the following text in the file `/home/hq6/vimwiki/diary/2019-03-20.wiki`.
```
= First Level Header =
This is the diary entry.

= First Level Header 2 =
This is the diary entry.

== Second Level header ==
How about something here
```

2. Open `/home/hq6/vimwiki/diary/diary.wiki` and run `let g:diary_subcaption_level = 2`.

3. Run `VimwikiDiaryGenerateLinks`.

Output:

```

= Diary =

== 2019 ==

=== March ===
    * [[2019-03-20|First Level Header]]
```

I expected to see links to all three headers appear in the index page. Am I using it incorrectly?

edlanglois:
The default setting (diary_caption_level=0) preserves the existing behaviour before this change and just takes the first header. Setting diary_caption_level > 0 will show all headers up to that level (e.g. =1 will show the two top level headers only, 2-6 will show all three).

hq6:
@edlanglois , I ran `let g:diary_subcaption_level = 2` without any difference.

edlanglois:
Oh, the setting is now `diary_caption_level` not `diary_subcaption_level` (I'll edit my original message in this PR).

Also, I set all vimwiki options using the `vimwiki_list` variable, does `let g:<name>` work for other options?:

```vim
let g:vimwiki_list = [{
	\ 'diary_caption_level': 1,
	\}]
```

edlanglois:
It's been awhile so I had forgotten how the vimwiki options work. Looks like vimwiki_list is for wiki-local options and `g:vimiki_<name>` for global options. I don't have a strong preference where the caption level option goes but other comparable diary options (`diary_header`, `diary_sort`) are also wiki-local.

hq6:
I have manually merged this PR as 8f5d38365bcd73fc44847b4de5d9f4b10c64461f.

jessebett:
This does not appear to work with Markdown syntax vimwiki. I have set `'diary_caption_level':2` but if I have an entry with 
```
## Test

some stuff

## Test 2

other stuff
```

And then `:VimwikiDiaryGenerateLinks` the above behaviour does not happen, I only get the usual entry with `[Test](path/to/entry)`

edlanglois:
This feature has not been included in the `master` branch yet. I recommend checking out the `dev` branch if you want this feature but expect it to be less stable and possibly missing some other content from the master branch.

PR #500, opened 2018-05-25
Diary links should use markdown syntax when appropriate

### LINKED ISSUE #499: Links generated by `:VimwikiDiaryGenerateLinks` always use vimwiki syntax
When using markdown syntax, the generated links in the diary section still use the vimwiki/mediawiki syntax. The links still function as expected, but this mismatch causes some friction when using tools like pandoc. I'd expect the diary links to be consistent with the other linking behavior.

Thanks for keeping vimwiki alive and excellent!

hcwndbyw:
When markdown syntax is enabled, generated links should use markdown
link syntax.

Fix #499


EinfachToll:
Good work. Also, congratulation for claiming issue number 500 ;)

PR #512, opened 2018-06-09
Add a version that shows the version or commit

Nudin:
The Idea is to provide the user a simple command to get the current version number or if he cloned the dev-branch the revision id.
For a release `s:plugin_vers` would be set to the version number, for example "2.4", while it would be set to -1 in the dev-branch.

### DIFF
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -12,7 +12,7 @@ let g:loaded_vimwiki = 1
 " Set to version number for release, otherwise -1 for dev-branch
 let s:plugin_vers = -1
 
-" Get the plugin the script is installed in
+" Get the directory the script is installed in
 let s:plugin_dir = expand('<sfile>:p:h:h')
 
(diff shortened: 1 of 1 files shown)


plugin/vimwiki.vim
```diff
@@ -9,6 +9,11 @@ if exists("g:loaded_vimwiki") || &cp
 endif
 let g:loaded_vimwiki = 1
 
+" Set to version number for release, otherwise -1 for dev-branch
+let s:plugin_vers = -1
+
+" Get the plugin the script is installed in
```
EinfachToll: plugin -> directory

plugin/vimwiki.vim
```diff
@@ -186,6 +191,23 @@ function! s:set_windowlocal_options()
 endfunction
 
 
+function!  vimwiki#get_version()
```
EinfachToll: why not use s:get_version? Does this syntax even work? Also, there are two spaces.

PR #528, opened 2018-07-12
Add option to disable creating new wiki pages

sstallion:
This PR adds a new global option named `create_link`, which allows the user to control `:VimwikiFollowLink` behavior when attempting to follow a non-existent link. The original behavior of creating a new link is preserved, however by setting this option to `0`, new links will not be created when pressing return over regular text.


hq6:
@sstallion , What problem are you trying to address with this PR? 

Under today's behavior, if you create a new page by following a non-existent link, that page gets written if and only if you actually write content to the new page. Otherwise, there is no harm done, and nothing is persisted.

CrossR:
I'm interested in this since I'll occasionally catch the enter key and end up with `[word](word)` in my wiki pages. Being able to disable that and have enter only follow a link would be great.

The page itself isn't made or persisted, but I do still end up with random `[]()` links, when really I don't mind just manually typing out my links since I usually have a folder structure and a desired location to put a file.

sstallion:
That's exactly what led me to creating the patch. If I'm using a pop up
while editing a wiki page, I've occasionally had the stray return create a
new link, which is annoying to back out.

ranebrown:
I've had this happen to me a few times as well, usually when trying to follow a link when the cursor isn't actually on the link. Since the default behavior remains the same and this is a user option I think it is ok to merge.

@hq6 what do you think?

@sstallion  this will need updated to be against the `dev` branch before we can merge it.

hq6:
I concur.

sstallion:
Will do. I should be able to get this (and others) ready to go
today/tomorrow.

sstallion:
@ranebrown This PR should be ready to go, sorry for the delay!

hq6:
I have manually merged this to `dev` to clean up commit history.

ranebrown:
@hq6 No worries, having a cleaner commit history is usually a good idea.

@sstallion Thanks for the PR!

PR #530, opened 2018-07-12
Add option to automatically generate diary index

sstallion:
This PR adds an option to automatically generate the diary index page when opened. By default, existing behavior is preserved.

### DIFF
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -231,6 +231,11 @@ function! vimwiki#diary#goto_diary_index(wnum)
   endif
 
   call vimwiki#base#edit_file('e', s:diary_index(idx), '')
+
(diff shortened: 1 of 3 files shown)


zimbatm:
The auto_diary_index feature in this PR happens when visiting the diary index page.

In my case I don't think that I will be visiting that page a lot but still want to keep it fresh. It would be great if this could also happens every time a new diary page is also created.

Nudin:
Hi,
this PR seems good! But could you please change the destination branch to the `dev`-branch? Then I would merge it.
Thanks a lot for the contribution and sorry for the long delay – there are not many vimwiki-devs…

sstallion:
Sure thing! I apologize for the delay, I've been away on travel. I'll
redirect the PR momentarily.

Issue #406, opened 2017-10-28
Renaming file does not work for markdown syntax


yangkev:
When I create a link by pressing `<CR>`, and then follow that link, if I change the name of the file with `<leader>wr`, the index file's link does not get updated.

Info:
- Using Markdown

EinfachToll:
I allowed myself to change the title. Looks like renaming doesn't work at all for markdown syntax.

Nudin:
Can you check with latest version from dev-branch is the problem still exist? I just fixed #550 and this could be the same bug.

Nudin:
Fixed one more bug and tested it, seems to work now!?

ranebrown:
@yangkev This should be fixed. If you are still having issues pease reopen.

ranebrown:
This actually still seems to be an issue but lets track it in the other duplicate issue.

@Nudin Can you take a look at this since it is something you worked on in the past?

PR #533, opened 2018-07-20
Update the links when renaming markdown wikis

raphaelfeng:
It fixes https://github.com/vimwiki/vimwiki/issues/406.

Two reasons whey the links in markdown wikis are not updated:
1. The markdown link pattern is wrong, should be `[]()` rather than `[][]`;
2. The logic of getting wiki local var cannot get correct wiki index as:
    a. the renamed file's buffer is removed and `%:p` would return empty
    b. the function that gets the wiki local var depends on buffer's `%:p`
    value to find the wiki it belongs to and it would always return -1
    and result in the default option values instead the user's option

The fix is
1. fix the markdown link pattern regex;
2. keep the renamed file's buffer open during the period of updating the link

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1320,11 +1320,22 @@ function! vimwiki#base#rename_link()
     execute ':update'
   endfor
 
+  " activate the renamed buffer
   execute ':b '.escape(cur_buffer[0], ' ')
 
-  " remove wiki buffers
+  " rename the buffer name to the new name
+  " execute ':file '.new_link
+
+  " leave the renamed buffer and remove the rest wiki buffers
+  " the reason we leave the renamed buffer is because the get_wikilocal
+  " function use the current file name to get the vars of current file's
+  " wiki, but after calling rename function, %:p still returns empty
+  " rather than the new full file path, so we just leave the renamed file's
+  " buffer so that %:p would be the renamed full file path
   for bitem in blist
-    execute 'bwipeout '.escape(bitem[0], ' ')
+    if !vimwiki#path#is_equal(bitem[0], cur_buffer[0])
+      execute 'bwipeout '.escape(bitem[0], ' ')
+    endif
   endfor
 
   let setting_more = &more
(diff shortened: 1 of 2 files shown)


raphaelfeng:
Can we merge this pull request? @Nudin

autoload/vimwiki/base.vim
```diff
@@ -1320,11 +1320,22 @@ function! vimwiki#base#rename_link()
     execute ':update'
   endfor
 
+  " activate the renamed buffer
   execute ':b '.escape(cur_buffer[0], ' ')
 
-  " remove wiki buffers
+  " rename the buffer name to the new name
+  " execute ':file '.new_link
```
Nudin: Is this commented out on purpose?

Nudin:
Hi,
sorry for the long delay. There are currently way to little vimwiki-devs…
I'm not familiar with this part of the code but looked into it. Two things I noticed:
- There is a commented out code-line (see review-note)
- vimwiki also supports links in the format `[][]` in markdown mode. I have no idea if anyone is using them. It might me worth a try to nevertheless just do the change here and see if it breaks someones code.

I haven't tried out the code. I assume you did!?

raphaelfeng:
> Hi,
> sorry for the long delay. There are currently way to little vimwiki-devs…
> I'm not familiar with this part of the code but looked into it. Two things I noticed:
> 
> * There is a commented out code-line (see review-note)
> * vimwiki also supports links in the format `[][]` in markdown mode. I have no idea if anyone is using them. It might me worth a try to nevertheless just do the change here and see if it breaks someones workflow.
> 
> I haven't tried out the code. I assume you did!?

Thanks for the reply and I have replied the review comment.

Yes, I have tested my change and it worked well.

Nudin:
> Thanks for the reply and I have replied the review comment.

Sorry, I don't see any answer!?

raphaelfeng:
> > Thanks for the reply and I have replied the review comment.
> 
> Sorry, I don't see any answer!?

Forgot to submit the comment, submitted now.

samcoveney:
Hi,
Is there any progress on this issue of updating the links with markdown syntax vimwiki? This isn't working for me with the most recent vimwiki.
Thanks

raphaelfeng:
@samcoveney Can you confirm my fix works for you? If yes, we can nag @Nudin to merge the fix.

samcoveney:
@raphaelfeng Yes I can confirm that your fix works, thank you very much! With your fix, I can rename both vimwiki and markdown syntax files with \wr, the file name is changed and the links are updated in other files. Doesn't seem to have broken anything. Thanks again

Nudin:
If you rebase it to the `dev`-branch, I'll merge it into that. Merging into the `master`-branch should better be done by @EinfachToll.

raphaelfeng:
@Nudin Another PR for the dev branch created at https://github.com/vimwiki/vimwiki/pull/593

Nudin:
Merged #593 into dev. Thanks everyone!

PR #557, opened 2018-09-13
Configurable lists

Nudin:
The hopefully last reimplementation of #390 and #479. The patch looks quite long but is actually way simpler and strait-forward than the previous implementation.

- Allows to configure the allowed bullet-characters for lists.
   - As discussed it is now a wikilocal config variable allowing
   - The variable does not anymore extend the list of chars but overwrites it. Allowing also to free characters from their meaning as bullet-char
- Allows to set the listsymbols for todo-lists on a per-wiki-level
   - Since I had to touch the code anyway I also fixed this other long annoying limitation:
   - You can specify `listsyms` and `listsym_rejected` on a per-wiki-level and not only globally. This allows to work together with friends who use other symbols than you in their wikis. I think they should have been wikilocal in the beginning.

Open questions:
- I'm not totally completely happy about having the list specific regexes now in `g:vimwiki_wikilocal_vars` while all other regexes are in `g:vimwiki_syntax_variables`. This makes it ugier to tell users to post there `g:vimwiki_wikilocal_vars` in bug-reports. Should we create another dict for them?
- Should we mark `g:vimwiki_listsyms` and `g:vimwiki_listsym_rejected` as deprecated to remove them in a few years?

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -886,8 +886,8 @@ function! s:process_tag_list(line, lists)
     let st_tag = '<li>'
     let chk = matchlist(a:line, a:rx_list)
     if !empty(chk) && len(chk[1]) > 0
-      let completion = index(vimwiki#vars#get_syntaxlocal('listsyms_list'), chk[1])
-      let n = len(vimwiki#vars#get_syntaxlocal('listsyms_list'))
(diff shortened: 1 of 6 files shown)


Nudin:
After so many rebases and rewrites I'm now merging it, we can still improve the code.

PR #559, opened 2018-09-14
Table can have last cell with the smallest size possible to avoid ugly wrapping

tinmarino:
Add an option to NOT expand the cell of the last column of a table.
This permits to view (a 2 columns) table on small screen even when set wrap

### DIFF
--- a/autoload/vimwiki/tbl.vim
+++ b/autoload/vimwiki/tbl.vim
@@ -278,6 +278,10 @@ function! s:get_aligned_rows(lnum, col1, col2)
     call add(cells, vimwiki#tbl#get_cells(row))
   endfor
   let max_lens = s:get_cell_max_lens(a:lnum, cells, startlnum)
+  if vimwiki#vars#get_global('table_reduce_last_col')
+    let last_index = keys(max_lens)[-1]
+    let max_lens[last_index] = 1
+  endif
   let result = []
   for [lnum, row] in rows
     if s:is_separator(row)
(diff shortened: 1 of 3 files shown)


ranebrown:
I'm not convinced this is a feature that many users would find useful. There is a pretty big trade off since the table ends up being misaligned and it is pretty easy to just `set nowrap` and then scroll to view the table.

@tinmarino maybe you can provide a real life example of a table and the column width of the terminal you are viewing it in?

tinmarino:
Hi Rane, Thank for your concern.

I personally use (a lot) Vimwiki on my __Android__ (Termux), screen size is 40 columns. It is very useful to take notes and read on the road and then tidy all that at home (syncing with git).

vimwiki_no_reduce

vimwiki_reduce

tinmarino:
The first image without reducing last column and the second image reducing the last column removing useless whitespaces
I don't like to `set nowrap`. Because then I can't see long lines on my device.

Cheers and thanks for your work.

ranebrown:
Your images both look like the setting to not expand the column is applied? Regardless I was able to run a test with the additional information you provided. Since the default behavior remains the same and this is a pretty minimal change that might help some users I'm willing to merge it.

Here are a couple of requests:
1. Update this pull request to be against the `dev` branch. (See [CONTRIBUTING.md](https://github.com/vimwiki/vimwiki/blob/dev/CONTRIBUTING.md)).
2. Update the documentation to use the same format as the other commands with a `Value` and `Description` column.
3. Modify the wording "This permits to see better the table when |'wrap'|" to something like "When `:set wrap` this option improves how a table is displayed, particularly on small screens.

tinmarino:
The first image is expanding.
Thank for accepting merge (usefull for me at least).
I stay at your disposition if you need some help on any debug | merge | develop stuff.

tinmarino:
I Created a new pull request (cleaner and up to date) : #637
You can close this one.

PR #562, opened 2018-09-17
Don't include interwiki prefix in normalized link

farnerup:
Make `[[wiki1:foo]]` normalize to `[[wiki1:foo|foo]]` rather than `[[wiki1:foo|wiki1:foo]]`

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1858,6 +1858,11 @@ endfunction
 
 
 function! s:clean_url(url)
+  let wikiPattern = '^wiki[0-9]\+:'
+  if match(a:url, wikiPattern) == 0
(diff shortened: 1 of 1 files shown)


Nudin:
Thanks for the patch! Seems a good improvement. But if we strip the ´wikiN:´ prefix, we might also want to strip the `diary:`-prefix. Can you add this? And can you change the destination to the `dev`-branch?

Nudin:
Oh, and while you're on it: I just noticed that we remove `file:`-prefix but don't remove the `local:`-prefix. Could you fix that as well?
Thanks a lot in advance!

farnerup:
The problem is the lack of slashes after `local:`
Replaced with #563

PR #563, opened 2018-09-18
Clean URLs that don't have slashes

farnerup:
Will fix normalization of links like [[local:./foo.txt]]
Also, allow schemes to end with a number, so that e.g. [[wiki1:foo]]
will normalize as expected.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1861,7 +1861,8 @@ endfunction
 
 function! s:clean_url(url)
   " remove protocol and tld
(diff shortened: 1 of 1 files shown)


farnerup:
Replaces #562

Nudin:
I m not totally sure about this. With this all forms of `[[foo:bar]]` will be normalized of `[[bar]]` even if `foo` has no special meaning. But colons are rather rare in filenames, so I'll merge this for now to the dev-branch. We'll see if it breaks anything for anyone.

PR #573, opened 2018-10-18
Add %wiki_path% template variable.

HugoNikanor:
I needed a %wiki_path% variable for a git integration script I'm writing. I also thought that up-streaming it would be a good idea.

Recreated issue to point at different branch in my fork.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1533,6 +1533,7 @@ function! s:convert_file(path_html, wikifile)
 
     let title = s:process_title(placeholders, fnamemodify(a:wikifile, ":t:r"))
     let date = s:process_date(placeholders, strftime('%Y-%m-%d'))
+    let wiki_path = strpart(s:current_wiki_file, strlen(vimwiki#vars#get_wikilocal('path')))
 
     let html_lines = s:get_html_template(template_name)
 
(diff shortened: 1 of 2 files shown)


EinfachToll:
Ok, this may be useful for some users. If you would add a remark about it to the "Placeholders" section of the doc, this would be even greater.

HugoNikanor:
I added a short piece of documentation. See the second commit.

HugoNikanor:
Ping. Will this be merged?

hq6:
@HugoNikanor , I am a new maintainer and I'm interested in getting this merged. 

It seems that all the other placeholder variables start with a `%` but do not end with a `%`. Is there a good reason for `%wiki_path%` to break this convention?

HugoNikanor:
@hq6 As far as I can see the placeholder variables use two `%` characters in the HTML template file (default.tpl), and only a single leading `%` from inside the .wiki files.

hq6:
@HugoNikanor , In that case, should `doc/vimwiki.txt` have only a leading `%`?

This current PR shows `%wiki_path%` in that file.

hq6:
Sorry I think I just realized why I was confused.  If my new understanding is correct, `%wiki_path%` **is only meant to be used in the HTML template** files and never in the `.wiki` files. However, everything else in the Placeholders documentation is referring to directives placed **inside wiki files**.

Please correct me if my new understanding is wrong.

If I have understood this correctly, I think it would be more clear to document this in the description of `vimwiki-option-template_path` instead, as a section of "Additional Template Variables", and leave it out of Placeholders.  I realize that @EinfachToll  asked you to put it in Placeholders, but I hope both of you can understand why this might cause confusion.

@HugoNikanor, what are your thoughts on this?

HugoNikanor:
@hq6. I agree with you on the confusion. I submitted a new commit which moves the documentation.

PR #574, opened 2018-10-19
Fixed system file handlers for windows

### LINKED ISSUE #560: Error trying to open a file with spaces on Windows
# Problem
## Situation
You're trying to link to a local file with markdown syntax. When you press enter you want it to open with the correct program. This is how you would do it:

```
Check out this [image](file:cutecat.png) of my cat!
```

This works fine, except on Windows if the filename or the path leading up to it (including the path that this vimwikis base folder is at) has any spaces. In that case Windows can't parse the path correctly and spits out this error:

capture

# Solution
Take a look at the function responsible for opening files on Windows: https://github.com/vimwiki/vimwiki/blob/f55ec31675e372e2f59d51322b445ea91191ec2b/autoload/vimwiki/base.vim#L221
You check in the beginning for the +shellslash feature and create a new url specifically with it disabled.
```
    if exists('+shellslash')
      let old_ssl = &shellslash
      set noshellslash
      let url = shellescape(a:url, 1)
      let &shellslash = old_ssl
    else
      let url = shellescape(a:url, 1)
    endif
```
But you never use that new url, you keep using `a:url`, which is the one that was initially passed to the function.
```
    if &l:shell ==? "powershell"
      execute 'silent ! start ' . a:url
    else
      execute 'silent ! start "Title" /B ' . a:url
    endif
```
This `a:url` was never modified by [shellescape()](http://vimdoc.sourceforge.net/htmldoc/eval.html#shellescape()), which puts double quotes `"` around the file path, fixing the problem. So changing lines 235 and 237 to `url`, which I assume was the plan anyway, is the solution.

## Problem with the solution

**But wait**, this only works for cmd. Powershell still doesn't parse the file path correctly:

capture2

Powershell's filepaths appearantly need to be enclosed in single quotes `'` instead. You could achieve that by setting shellslash and running shellescape(), but refer to the top of the function where there is a comment describing that this is problematic. Now, that comment is from 2014 and maybe it's only problematic for cmd, and for powershell it's actually the correct solution, but I'm just guessing at this point.

Anybody with more knowledge about this might like to take a closer look.

Otherwise, I'll take a crack at it and create a pull request with changes to the logic as described:
For cmd: set noshellslash and run a:url through shellescape()
For powershell: set shellslash and run a:url through shellescape()

BenjaminBrandtner:
Fixes #560
Opening external files containing spaces should now work on cmd and powershell.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -220,22 +220,32 @@ endfunction
 function! vimwiki#base#system_open_link(url)
   " handlers
   function! s:win32_handler(url)
-    "http://vim.wikia.com/wiki/Opening_current_Vim_file_in_your_Windows_browser
-    "disable 'shellslash', otherwise the url will be enclosed in single quotes,
-    "which is problematic
-    "see https://github.com/vimwiki/vimwiki/issues/54#issuecomment-48011289
-    if exists('+shellslash')
-      let old_ssl = &shellslash
-      set noshellslash
-      let url = shellescape(a:url, 1)
-      let &shellslash = old_ssl
+    "Disable shellslash for cmd and command.com, but enable for all other shells
+    "See Issue #560
+    if (&shell =~? "cmd") || (&shell =~? "command.com")
+
+      if exists('+shellslash')
+        let old_ssl = &shellslash
+        set noshellslash
+        let url = shellescape(a:url, 1)
+        let &shellslash = old_ssl
+      else
+        let url = shellescape(a:url, 1)
+      endif
+      execute 'silent ! start "Title" /B ' . url
+
     else
-      let url = shellescape(a:url, 1)
-    endif
-    if &l:shell ==? "powershell"
-      execute 'silent ! start ' . a:url
-    else
-      execute 'silent ! start "Title" /B ' . a:url
+
+      if exists('+shellslash')
+        let old_ssl = &shellslash
+        set shellslash
+        let url = shellescape(a:url, 1)
+        let &shellslash = old_ssl
+      else
+        let url = shellescape(a:url, 1)
+      endif
+      execute 'silent ! start ' . url
+
     endif
   endfunction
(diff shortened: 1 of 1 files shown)


autoload/vimwiki/base.vim
```diff
@@ -220,22 +220,32 @@ endfunction
...
-      let url = shellescape(a:url, 1)
-      let &shellslash = old_ssl
-    else
-      let url = shellescape(a:url, 1)
-    endif
-    if &l:shell ==? "powershell"
-      execute 'silent ! start ' . a:url
+    "Disable shellslash for cmd and command.com, but enable for all other shells
+    "See Issue #560
+    if (&l:shell =~ "cmd") || (&l:shell =~ "command.com")
```
EinfachToll: I know you didn't introduce it, but can you please remove the `l:`, while we are at it?
Also, never use `=~` but always either `=~#` or `=~?`, else the behavior of this command may depend on the user's 'ignorecase' setting which is a nightmare to debug. (Although it's probably not a big problem in our case here.)
```suggestion
    if (&shell =~# "cmd") || (&shell =~# "command.com")
```

BenjaminBrandtner:
I did something wrong. I implemented your suggested changes, but as I didn't think it needed it's own commit I ammended the previous one and force pushed it to my forked repository.
But that didn't update this pull request, it still wants to merge the now outdated commit. 
Should I open a new PR?

BenjaminBrandtner:
Help.. this PR is bugging out, presumably because I force-pushed the changes you proposed to my fork. Now it keeps changing which commit it's supposed to merge.

BenjaminBrandtner:
[Ah, actually it was all of Github that was bugging out.](https://blog.github.com/2018-10-21-october21-incident-report/) I wasn't at fault. Hooray! 
Well, I'll have to open a new PR, as I deleted my fork trying to fix things. Sorry about the mess!

Issue #585, opened 2018-12-12
Anchor links broken for multiple extension (eg .wiki.txt)


nicolasbrailo:
Creating an anchor link (eg a link to a header) doesn't work when a wiki has multiple extensions (such as .wiki.txt). Example:

.vimrc
```
let testwiki = {}
let testwiki.path = '~/tmp'
let testwiki.ext = '.foo.bar'
```

~/tmp/index.foo.bar
```
[[#Test|link1]]
[[index#Test|link2]]

= Test =
```

When following link1, vimwiki will try to open ~/tmp/index.foo.foo.bar#Test. link2 works as expected, moving the cursor to the header "Test".

I managed to trace this to vimwiki#base#resolve_link, where fnamemodify(source_file, ':p:t:r') will delete only the last file extension (as per http://vimdoc.sourceforge.net/htmldoc/cmdline.html#filename-modifiers).

PR #586, opened 2018-12-12
Patch for issue #585, fix anchor links for multiple extensions

nicolasbrailo:


Nudin:
Thanks! I don't have time to review right now. (It's not much code but these things can have tricky side-effects!) Hope to manage to look at it around Christmas. If anyone else want's to test it that would be awesome.

nicolasbrailo:
Indeed, a thorough code review before accepting this PR is a good idea. While from my point of view the behavior looks like a bug, I'm sure it's possible it will break someone's workflow (https://xkcd.com/1172/, right?)

We can also choose to make this change somewhat backwards compatible by uncommenting line 162 - I chose not to because the current behavior seems incorrect, but I'd love the opinion of someone else on that.

autoload/vimwiki/base.vim
```diff
@@ -152,7 +152,16 @@ function! vimwiki#base#resolve_link(link_text, ...)
       let link_infos.anchor = join(split_lnk[1:], '#')
     endif
     if link_text == ''  " because the link was of the form '#anchor'
-      let link_text = fnamemodify(source_file, ':p:t:r')
+      if source_file =~ vimwiki#vars#get_wikilocal('ext') . '$'
```
EinfachToll: never use `=~`, instead use either `=~#` or `=~?`, otherwise it depends on the user's settings whether case matches or not. Here, `=~#` is correct.

autoload/vimwiki/base.vim
```diff
@@ -152,7 +152,16 @@ function! vimwiki#base#resolve_link(link_text, ...)
...
+          " Source file has expected extension. Remove it, it will be added later on
+          let ext_len = strlen(vimwiki#vars#get_wikilocal('ext'))
+          let link_text = fnamemodify(source_file, ':p:t')[:-ext_len-1]
+      else
+          " Source file has unexpected (or no) extension. A new file with the proper
+          " extension appended will be created
+          " let link_text = fnamemodify(source_file, ':p:t:r')
```
EinfachToll: Remove the whole `else` part. If we end up in this part of the code, something is clearly wrong, but I don't know how one would handle it here anyway.

autoload/vimwiki/base.vim
EinfachToll: Also, when a custom string is used as a regular expression, some characters should be escaped in the string. You can use `vimwiki#u#escape()`.

nicolasbrailo: Done! I had no idea about the #/? modifiers, thanks!

How will this behave in windows? A windows user may expect the match to be case insensitive. Is =~# still the right choice? I don't have a windows setup where I can test this though.

autoload/vimwiki/base.vim
nicolasbrailo: Done. If we keep the else branch (and uncomment line 162) we can preserve the old (buggy?) behavior. I'm not sure if anyone would expect the old behavior though, so I'm more than happy to completely remove it!

autoload/vimwiki/base.vim
EinfachToll: That's a valid point actually, and there is already a function for that (`vimwiki#path#is_equal()`), but I don't think we should generally allow mixed case in file extensions on windows, so I think this is OK.

Issue #179, opened 2015-12-25
Renaming wiki files in a subdirectory


d80harri:
Date seen: 2015-12-25
Versions:
- Win7/cygwin
- vim-7.4

Bug Description:
Links to files in subfolders are not renamed.

Steps to reproduce:
1. $ vim test.wiki
2. Create a link ("[[subdir/index]]")
3. Move cursor to link in normal mode and press enter
4. Vim asks if a new folder should be created. Answer with 'y'
5. :w to store the newly created (empty) wiki page
6. <LEADER>wr to rename
7. vim asks if renaming the file is ok; answer with 'y'
8. enter a new name (eg: "index2")
9. Hit Backspace to return to test.wiki

Actual Behavior
The link has not changed from subdir/index to subdir/index2

Expected Behavior
The link should now be [[subdir/index2]]

I do not know how vimwiki knows which files link to the actual page, but executing a ":VimwikiBacklinks" on "subidr/index" before the rename returns with "/home/hbi/Test.wiki". So if vimwiki knows the backlink it should change it, shouldn't it?

Cheers,

Harald

Michael-F-Ellis:
Reproducible in recent clone (Jan 30, 2017) on OS X 10.12.3, Vim 7.4

Love to see this one fixed.  That aside, Vimwiki is awesome. I've been using Vim for 20 years and this is as close to life-changing as it gets.  I've been _really_ wanting a text-based organizer and even tried switching to Emacs just for org-mode but it wasn't worth the pain that came with the rest of Emacs.

Thanks much for developing and maintaining Vimwiki.

ranebrown:
The subdirectory issue seems to be fixed on the `dev` branch though there are still some related issues such as `[[../page]]` links and markdown links not being renamed. (#237 #90)

PR #591, opened 2018-12-21
Fixed updation of links when renaming wiki files in subdirectories

metastance:
This solves issue #179


Nudin:
Thanks a lot! :+1:

PR #593, opened 2018-12-27
Update the links when renaming markdown wikis

raphaelfeng:
Two reasons whey the links in markdown wikis are not updated:
1. The markdown link pattern is wrong, should be []() rather than [][];
2. The logic of getting wiki local var cannot get correct wiki index as:
    a. the renamed file's buffer is removed and %:p would return empty
    b. the function that gets the wiki local var depends on buffer's %:p
    value to find the wiki it belongs to and it would always return -1
    and result in the default option values instead the user's option

The fix is
1. fix the markdown link pattern regex;
2. keep the renamed file's buffer open during the period of updating the link

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1322,11 +1322,22 @@ function! vimwiki#base#rename_link()
     execute ':update'
   endfor
 
+  " activate the renamed buffer
   execute ':b '.escape(cur_buffer[0], ' ')
 
-  " remove wiki buffers
+  " rename the buffer name to the new name
+  " execute ':file '.new_link
+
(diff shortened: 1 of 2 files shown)


EinfachToll:
I appreciate the effort, but I reverted the commit, because it breaks the wiki files for every Vimwiki user who uses markdown syntax. Can't off-handedly change the syntax, folks! I'm all for improving markdown support, because apparently, many people use it, but we have to keep it reasonably backwards compatible, for example by adding (instead of replacing) the `[foo](bar)` syntax and changing the `[foo][bar]` syntax (which indeed exists in markdown, but has different semantics) in a backwards-compatible way.

PR #594, opened 2018-12-27
Minor doc fixes

fte10kso:


Nudin:
Hi,

Thanks for the PR! But I think that the original version was right – even through confusing. What do you think about changing it the following way:

> It is possible to toggle several list items using visual mode. But note that
> instead of toggling every item individually, all items get checked if the
> first item **was** unchecked and all items get unchecked if the first item **was**
> checked.

fte10kso:
Hey, I must have totally misread that sentence. I guess I was still stuck on nested lists from the previous section interpreting *first* as *parent*.

I've changed *is* to *was* according to your suggestion instead.

Nudin:
Thanks!

PR #599, opened 2019-01-05
Made headings link to themselves.

HugoNikanor:
Made each heading link to itself, by placing an <a> tag inside each
<h[1-6]> tag.

Also adds class="header" to each header, allowing for the CSS rule
    .header a { ...  }
to remove all styling from the inner links, allowing the outputed
document to look like before.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1072,7 +1072,9 @@ function! s:process_tag_h(line, id)
         let h_text = num.' '.h_text
       endif
       let h_complete_id = s:escape_html_attribute(h_complete_id)
-      let h_part = '<div id="'.h_complete_id.'"><h'.h_level.' id="'.h_id.'"'
+      let h_part  = '<div id="'.h_complete_id.'">'
+      let h_part .= '<h'.h_level.' id="'.h_id.'" class="header">'
+      let h_part .= '<a href="#'.h_complete_id.'"'
 
     else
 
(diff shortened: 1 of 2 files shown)


EinfachToll:
I don't get the use case. Why would somebody want a header to link to itself? Are there examples of websites etc. where this is done?

HugoNikanor:
The use case is easier linking to a part of the webpage. So
if you want to highlight a specific section this allows you
to better do that.

One example of a page that has this is (which I just found
by looking at reddit.com/r/programming for 10s):
https://www.jering.tech/articles/semantic-versioning-in-practice.
When you hover over any heading a link icon appears to the
right, giving you the same type of link that this patch
creates. 

-- 
Hugo Hörnquist

EinfachToll:
Oh, ok. Of course I have seen this before, but I was just confused why you made the whole header a link. So your patch with the headings being links instead of small buttons beside the header is supposed to be a poor man's solution, right? However, I would rather prefer something which looks halfway good out-of-the-box, and blue underlined headers are just confusing.
So, if you remove the link creation from your patch, I would merge it (or if you provide a solution which looks like the site you mentioned).

HugoNikanor:
It is indeed a poor mans implementation of it. The patch to the default CSS should remove the blue underlines.

> if you remove the link creation from your patch

What do you mean here? My whole patch is link creation.

---

It's also possibly to do something like

    .header a:hover::after {
        content: "link-symbol";
    }

to indicate that it's a link. But you need CSS for the copy on click function. Which I believe we both want to avoid.

EinfachToll:
Whoops, never mind! While testing out your changes, I used the old style.css which drew the headers as blue underlined links, therefore my skepicism. Making the headers links without changing the visuals is indeed a good solution, thanks!

PR #601, opened 2019-01-07
Enhance fold handling

### LINKED ISSUE #580: Vim gets slow when edditing long wiki files
In the last few days I noticed Vim got slower, but I couldn't figure out why. Now I found that the main reason is because I updated this plugin and I maintaining a few long vimwiki files open (~10k lines, ~150 level 2 headers).

For example, when switching to a tabpage it may take a long time (~6 seconds) to finish the redraw if the cursor will end on a wiki page of the new tab. Removing the [`BufEnter` autocmd that calls `s:setup_buffer_enter()`](https://github.com/vimwiki/vimwiki/blob/dev/plugin/vimwiki.vim#L236) solves the issue.

I also found that editing the long wiki files is slow, even though the fix for #225 is in place. For example, inserting  a newline on the middle of text with normal mode <kbd>o</kbd> take ~6 seconds, while insert mode <kbd>Enter</kbd> takes ~12 seconds, and removing those two mappings makes things faster.

I have little knowledge of this cod, and I'm wondering if it is possible to optimize the implementation of those functions, or if it is better to disable them only for long files.

mMontu:
Setting `'foldmethod'` always causes all folds to be re-calculated, even though its is being written with same value currently set.

The  `'foldmethod'` was being reset every time the cursor entered a vimwiki, causing an uncessary large delay. So it was changed to set it only the first time the buffer is loaded into a window.

Fix #580

### DIFF
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -268,7 +268,10 @@ augroup vimwiki
     exe 'autocmd BufNewFile,BufRead *'.s:ext.' call s:setup_new_wiki_buffer()'
     exe 'autocmd BufEnter *'.s:ext.' call s:setup_buffer_enter()'
     exe 'autocmd BufLeave *'.s:ext.' call s:setup_buffer_leave()'
-    exe 'autocmd BufWinEnter,DiffUpdated *'.s:ext.' call s:setup_buffer_win_enter()'
+    exe 'autocmd BufWinEnter *'.s:ext.' call s:setup_buffer_win_enter()'
+    if exists('#DiffUpdated')
+      exe 'autocmd DiffUpdated *'.s:ext.' call s:setup_buffer_win_enter()'
+    endif
     " Format tables when exit from insert mode. Do not use textwidth to
     " autowrap tables.
     if vimwiki#vars#get_global('table_auto_fmt')
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -313,7 +313,7 @@ augroup vimwiki
     exe 'autocmd BufEnter *'.s:ext.' call s:setup_buffer_enter()'
     exe 'autocmd BufLeave *'.s:ext.' call s:setup_buffer_leave()'
     exe 'autocmd BufWinEnter *'.s:ext.' call s:setup_buffer_win_enter()'
-    if exists('#DiffUpdated')
+    if exists('##DiffUpdated')
       exe 'autocmd DiffUpdated *'.s:ext.' call s:setup_buffer_win_enter()'
     endif
     " automatically generate a level 1 header for new files
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -102,11 +102,21 @@ function! s:setup_buffer_enter()
     return
   endif
 
+  call s:set_global_options()
+endfunction
+
+
+" this is called when the buffer enters a window or when running a  diff
+function! s:setup_buffer_win_enter()
+  " don't do anything if it's not managed by Vimwiki (that is, when it's not in
+  " a registered wiki and not a temporary wiki)
+  if vimwiki#vars#get_bufferlocal('wiki_nr') == -1
+    return
+  endif
+
   if &filetype != 'vimwiki'
     setfiletype vimwiki
   endif
-
-  call s:set_global_options()
 
   call s:set_windowlocal_options()
 endfunction
@@ -258,6 +268,7 @@ augroup vimwiki
     exe 'autocmd BufNewFile,BufRead *'.s:ext.' call s:setup_new_wiki_buffer()'
     exe 'autocmd BufEnter *'.s:ext.' call s:setup_buffer_enter()'
     exe 'autocmd BufLeave *'.s:ext.' call s:setup_buffer_leave()'
+    exe 'autocmd BufWinEnter,DiffUpdated *'.s:ext.' call s:setup_buffer_win_enter()'
     " Format tables when exit from insert mode. Do not use textwidth to
     " autowrap tables.
     if vimwiki#vars#get_global('table_auto_fmt')
--- a/syntax/vimwiki.vim
+++ b/syntax/vimwiki.vim
@@ -488,6 +488,7 @@ call vimwiki#base#nested_syntax('tex',
 
 syntax spell toplevel
 
-syntax sync maxlines=300
+syntax sync maxlines=300  " speed up screen update by limiting the number of lines parsed
+" 'grouphere' used to speed up the syntax highlight
 syn sync match VimwikiSyncPat	grouphere  VimwikiH1Folding	"^\s*="
 syn sync match VimwikiSyncPat	grouphere  VimwikiH1Folding	"}}}"
--- a/syntax/vimwiki.vim
+++ b/syntax/vimwiki.vim
@@ -488,3 +488,6 @@ call vimwiki#base#nested_syntax('tex',
 
 syntax spell toplevel
 
+syntax sync maxlines=300
+syn sync match VimwikiSyncPat	grouphere  VimwikiH1Folding	"^\s*="
+syn sync match VimwikiSyncPat	grouphere  VimwikiH1Folding	"}}}"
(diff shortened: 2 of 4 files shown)


hq6:
Please checkout the `dev` branch and make this PR against the `dev` branch.

hq6:
Since there has been no response on this, I am closing this PR for repo sanitation.

If there is renewed interest in doing this, please ping on this thread or create a new PR.

mMontu:
It is strange, the PR is ignored for more than two months but it is closed for lack of response in six days.
Anyway, I'm going to port it to dev as soon as possible.

But last time I checked the dev branch was not in use. @hq6 do you have more information about how/which branches are in use?

hq6:
@mMontu, There was a long period where the only active maintainer did not have enough time to handle the torrent of Pull Requests and Issues. A little over a month ago, @ranebrown  and I were appointed as new maintainers and we have been working through the backlog to try to put the repository into a maintainable state, so that PR's can be addressed in days instead of months.

The `dev` branch is where all pull requests are merged. Maintainers and adventurous users use the `dev` branch and report issues and regressions. If dev appears stable and accumulates a bunch of features, we will do a release and merge `dev` into `master`.

Since you have now responded to this PR, I am reopening it..

ranebrown:
@mMontu when you update this please provide some profiling of the speed differences between the current behavior and your update.

mMontu:
@hq6 thanks for the clarification!

@ranebrown please check the profile results below.
It can be reproduce with [this file](https://github.com/vimwiki/vimwiki/files/2590344/delay.zip) by running the following:

    $ gvim -u NONE -U NONE -N -i NONE
    <commands below can be copied to clipboard and run with `:@+`>
    set nocompatible
    filetype plugin on
    syntax on
    let g:vimwiki_camel_case = 0  " Don't make links from CamelCased words
    let g:vimwiki_table_auto_fmt = 0 " Turn off table auto-formatting
    let g:vimwiki_autowriteall = 0
    let g:vimwiki_folding = 'expr:quick'    " Enable folding.
    let &rtp.=',~/.vim/bundle/vimwiki'
    source ~/.vim/bundle/vimwiki/plugin/vimwiki.vim
    e ~/tmp/delay.wiki
    tabe
    let start = reltime() | tabprev | redraw | echom reltimestr(reltime(start))

Results:
* original base: around 6s  (hash 7ffc295)
* original PR: 0.02s
* current dev: 5.9s 
* curretn PR: 0.03s

While I was testing the updated code I noticed that navigating between tabs with long wiki files is far slower than it was on master a few months ago (something about 500ms for some long wiki files).
Another fix will probably be required. 

Are there automated tests for this plugin? It'd be good to include the test above to avoid regressions.

syntax/vimwiki.vim
```diff
@@ -488,3 +488,6 @@ call vimwiki#base#nested_syntax('tex',
 
 syntax spell toplevel
 
+syntax sync maxlines=300
```
hq6: I may be revealing my ignorance here, but I had to look at the help pages to get an idea of what this was doing. Can you add a couple of comments explaining the purpose of these three lines?

plugin/vimwiki.vim
```diff
@@ -258,6 +268,10 @@ augroup vimwiki
     exe 'autocmd BufNewFile,BufRead *'.s:ext.' call s:setup_new_wiki_buffer()'
     exe 'autocmd BufEnter *'.s:ext.' call s:setup_buffer_enter()'
     exe 'autocmd BufLeave *'.s:ext.' call s:setup_buffer_leave()'
+    exe 'autocmd BufWinEnter *'.s:ext.' call s:setup_buffer_win_enter()'
+    if exists('#DiffUpdated')
+      exe 'autocmd DiffUpdated *'.s:ext.' call s:setup_buffer_win_enter()'
```
ranebrown: What is the use case for the DiffUpdated autocmd?

Nudin:
I checked the benchmark with neovim 0.2.2:
```
1.08 s - master
1.09 s - dev
0.007 s – this PR
```

@mMontu we don't have any tests yet, but are currently planing to start collecting them.

mMontu:
@hq6 done, please suggest any additional clarification.

@ranebrown the use case I found regarding `DiffUpdate` was that vimwiki folds were broken after a `:diffoff`.

@Nudin nice, your results shows an even greater gain with this PR. I guess the difference can be explained by my slower machine & gVim running on a VM.

mMontu:
> @mMontu we don't have any tests yet, but are currently planing to start collecting them.

Good, I'd happy to help when you start adding them.

syntax/vimwiki.vim
```diff
@@ -488,3 +488,7 @@ call vimwiki#base#nested_syntax('tex',
 
 syntax spell toplevel
 
+syntax sync maxlines=300  " speed up screen update by limiting the number of lines parsed
+" 'grouphere' used to speed up the syntax highlight
+syn sync match VimwikiSyncPat	grouphere  VimwikiH1Folding	"^\s*="
```
ranebrown: @mMontu I think it would makes sense to add a `grouphere` to match markdown headers as well. What do you think?

mMontu: @ranebrown I agree. In fact, I thought that this was being done by this line. Am I'm missing something?

ranebrown: I think that only covers the default syntax headers `= Header =` and not markdown syntax headers `# Header`.

[Documentation here](https://github.com/vimwiki/vimwiki/blob/6766c37ce09a623b3de7ddb86ca30278eea37a82/doc/vimwiki.txt#L1997-L2010)

mMontu: Oh, I wasn't aware that the markdown syntax was accepted in vimwiki. 

How do you suggest we do it? I don't think we could duplicate the approach on this line, since `#` is used as list item in the default vimwiki markdown.

ranebrown: I think it will work to change the regex pattern to `vimwiki#vars#get_syntaxlocal('rxH1')` instead of adding another line. The syntax will be defined based on the open wiki syntax so there shouldn't be a conflict. If that doesn't work for some reason it can go in `syntax/vimwiki_markdown_custom.vim`.

mMontu: @ranebrown I've never seen dynamic pattern on the syntax commands, I'm not sure that will work.

In any case, I believe further improvement of the syntax highlight could be handled on a separated PR. It seems it will require a more complicated approach, so in case something goes wrong it would allow reverting it without affecting the changes made here to the `plugin/vimwiki.vim`.

ranebrown: In that case I would say remove these lines altogether and address that part in a different PR rather than partially implementing something here. Is the performance improvement tied to this section at all or is it mostly related to the autocommand?

ranebrown:
> @mMontu we don't have any tests yet, but are currently planing to start collecting them.
Good, I'd happy to help when you start adding them.

If you want to create a top level `tests/` directory and add a test as part of this PR that would be great.

mMontu:
> In that case I would say remove these lines altogether and address that part in a different PR rather than partially implementing something here. Is the performance improvement tied to this section at all or is it mostly related to the autocommand?

@ranebrown , done, I repeated the profiling and didn't notice differences.

I created the test file as you suggested. I did a small change on it, to open a fold in the middle of the file, which seems to cause an increase on the delay.

ranebrown:
@mMontu I think the test for the `DiffUpdated` autocmd should actually be `if exists('##DiffUpdated')` since a single `#` only checks if an event is subscribed to that autocmd not if it is supported.

Can you also clean up the commit history? Good resource [here](https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history) or you can just do a `git reset --soft HEAD~6` and then commit everything as a single commit.

Thanks!

mMontu:
@ranebrown nice catch about `DiffUpdated` -- updated accordingly.

About the commit history: as the PR was created four months ago, changed its base, and solving conflicts from other PRs that were merged, now its commits are mixed with commits from other people.
It should be easier to use the 'squash and merge' on the repo.

PR #609, opened 2019-02-11
Allow AddHeaderLevel and RemoveHeaderLevel to take counts, Fix  #608

### LINKED ISSUE #608: Cannot use `=` command with a count
If I use `=` to increase the header level, it increases by 1. If I use `4=` to increase the header level, it still increases by 1, rather than 4.

I think it would be better if `4=` increased the header by 4.

hq6:
This PR fixes #608.


ranebrown:
@hq6 I like the fix for `=` and `-` to take a count! Works for default syntax and Markdown.

The `tt` mapping is nice but I think we might need to discuss if this is the best key combination since it doesn't match most of the other `gl` keybindings. I also noticed if there are no unfinished tasks it prints the message: `Vimwiki: Wiki link not found` which isn't the most helpful for this use case.

Not sure if you want to just close this  and work directly on the `dev` branch now or update the pull request to be against `dev`?

hq6:
@ranebrown , I will update this PR to be against `dev`, which I hadn't known about when I created this PR. Also, the `tt` mapping was supposed to be a separate PR. I'll do that separately and we can figure out what the best key combination is.

PR #616, opened 2019-02-19
HTML color colorcodes.

HugoNikanor:
This colors all CSS color codes found alone in inline code blocks to the color they mention.
So `#00cc00` Would show up as green, but `_#00cc00` would show up as normal (probably gray).

Two commits are included, one simply changes the background color, while the other also calculates the inverse of the color and sets that as the text color.

I have set up a demo at http://wiki.hornquist.se/Colors.html.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -349,8 +349,12 @@ function! s:tag_code(value)
   let l:str = s:mid(a:value, 1)
   let l:match = match(l:str, '^#[a-fA-F0-9]\{6\}$')
 
+
   if l:match != -1
-    let l:retstr .= " style='background-color: " . l:str . ";'"
+    let l:inv_color = 0xFFFFFF - eval("0x" . l:str[1:])
+
+    let l:retstr .= printf(" style='background-color: %s; color: #%x;'",
+          \ l:str, l:inv_color)
   endif
 
   let l:retstr .= '>'.s:safe_html_preformatted(l:str).'</code>'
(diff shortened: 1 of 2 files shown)


hq6:
@Nudin @ranebrown , Since this is new feature which has not yet received the attention of other maintainers, I'd like to get your thoughts whether to merge it.

ranebrown:
My initial thought is why not use something like [colorizer](https://github.com/lilydjwg/colorizer) that also works for files outside of vimwiki?

hq6:
@ranebrown , I think this PR is not about syntax highlighting *inside* Vim. It's about the generated HTML from `Vimwiki2HTML`, which does not overlap with colorizer, as far as I can tell.

ranebrown:
Ah, that makes more sense. I haven't had a chance to test it myself but the changes seem pretty minimal. I'll give it a try this weekend but if there aren't any issues I think we should merge it.

HugoNikanor:
@ranebrown It's pretty much that plugin, but for the HTML output.

P.S. Thanks for showing me that plugin, I was looking for one!

Nudin:
I like the feature and the code looks good. Was first a bit hesitated if that might cause a regression for someone since not all hex-codes might be colors. But I'm not sure how big the chance is that someone uses them after a `#` in a inline code block without anything else. Instead of using the inverse color for the text we might want to use black or white depending on the luminosity of the color. See [here](https://stackoverflow.com/a/1855903). We might also want to document this feature with a sort sentence in the help-page.

ranebrown:
I gave this a spin and it seems to work as advertised. I agree that it needs a better scheme for determining the font color. A few colors that the font is difficult to see are `#808000` and `#808080` and `#ffe119`. I'm sure there are plenty of other examples as well.

@HugoNikanor can you update this to use something like @Nudin suggested for determining the font color and add a short note in the documentation? Thanks!

HugoNikanor:
Fixed foreground colors and added documentation, and also updated the preview at http://wiki.hornquist.se/Colors.html.

(I also see that GitHub has added the same functionality...)

PR #627, opened 2019-03-13
Allow customization of header level of TOC

### LINKED ISSUE #626: TOC can only be at header level 1
It would be nice to have this be customizable so that a user could specify that they want the ToC at a different header level.

gpanders:
Instead of forcing the TOC to always be at header level 1, allow the
user to specify via the option g:vimwiki_toc_header_level what level
they want.

This defaults to 1, so if the user does nothing then the old behavior
will remain the same.

Closes #626.


hq6:
Please checkout the `dev` branch and make this PR against the `dev` branch.

gpanders:
I'll have to re-create the PR to change the source branch. Closing this one.

PR #629, opened 2019-03-14
Allow customization of header level of TOC

### LINKED ISSUE #626: TOC can only be at header level 1

gpanders:
Instead of forcing the TOC to always be at header level 1, allow the
user to specify via the option g:vimwiki_toc_header_level what level
they want.

This defaults to 1, so if the user does nothing then the old behavior
will remain the same.

Closes #626.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -358,7 +358,7 @@ function! vimwiki#base#generate_links()
 
   let links_rx = '\m^\s*'.vimwiki#u#escape(vimwiki#lst#default_symbol()).' '
 
-  call vimwiki#base#update_listing_in_buffer(lines, 'Generated Links', links_rx, line('$')+1, 1)
+  call vimwiki#base#update_listing_in_buffer(lines, 'Generated Links', links_rx, line('$')+1, 1, 1)
 endfunction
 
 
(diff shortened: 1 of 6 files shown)


hq6:
For posterity, can you provide the files you tested this with, and the procedures you used to test?

gpanders:
`~/.vimrc`:
```vim
let g:vimwiki_toc_header_level = 3
```

In a new file called `index.wiki`, write the following:
```wiki
= Foo =
    * Test

== Bar ==
    * Test
```

Run `:VimwikiTOC`. Expected output should be:
```wiki
=== Contents ===
    - [[#Foo|Foo]]
        - [[#Foo#Bar|Bar]]
```

Run `:Vimwiki2HTML`. Verify that the header under `<div id="Contents" class="toc">` is `<h3>`.

If `let g:vimwiki_toc_header_level` is not specified, then the expected output is
```wiki
= Contents =
    - [[#Foo|Foo]]
        - [[#Foo#Bar|Bar]]
```
as before, and the resulting HTML will have an `<h1>` header.

ranebrown:
I like this change. The only issue I see is that the TOC is still inserted at the top of the file. This most likely doesn't make sense if using a header level other than 1 (you probably don't want to end up with H3 for the TOC followed by an H1 header for something else).

It does work to manually move the TOC to a desired location and then updates work on the that location. Any ideas on a better way to handle this?

gpanders:
I think having the TOC inserted at the top by default is fine, since there is nothing that says an h3 or h2 cannot precede an h1. My motivation for doing this was actually because I wanted the TOC at the top, but I didn't want the "Contents" header to be the same size/boldface as my "actual" headers.

In the case that a user wants the TOC somewhere else in their document, they can manually initialize it wherever they want, as you said.

ranebrown:
Makes sense. I think it would be useful to have a command that inserts the TOC at the current cursor position (or something similar) but that can be left for another day. Thanks for the PR!

Issue #15, opened 2013-09-12
Proposal for Markdown as Default markup for Vimwiki files


Harsh-Vardhan:
Please don't get me wrong, I have learned the Vimwiki markup and it is burnt on my fingers so much that I find myself organizing my normal text files with it. However when I made a new wiki purposed to act as my blog I decided to go with the markdown syntax.

I resorted to markdown mainly because I wanted more features from the default **converter to HTML** in Vimwiki and have started learning it. **Markdown** has gathered much support as the writing system for the web. In addition parsers and converters to HTML for the **markdown** syntax are more mature and are being developed collaboratively by many.

If we stick with the custom markup originally developed for Vimwiki we will lose out on the collective intelligence of programmers working on markdown. By adopting a well-known and well-supported writing standard as default for Vimwiki I can see many benefits
- no documentation or support required for our original markup
- reduced barrier to entry for new users
- better support for parsers and converters to HTML
- one less feature to maintain and support

I hope you'd consider my proposal.

a-b:
It would be great to have github markdown https://guides.github.com/features/mastering-markdown/

it might be an amazing feature to publish diary or page as gist using gist plugin

EinfachToll:
@a-b I don't know if I get you right, but alternative syntaxes (e.g. markdown and media wiki) are already supported (see `:h vimwiki-option-syntax`), it's just that markdown support is still incomplete (e.g. links)

blessedbiped:
I think perhaps part of the problem is that it's not quite clear how to enable markdown syntax. Just got done reading the help file and I'm still stumped on how to make it happen. Should I just paste the example text into my vimrc, or is there another configuration file somewhere that I'm missing?

blessedbiped:
Figured it out, for anyone else looking for this option you **do** need to put the example text in your vimrc.

EinfachToll:
@blessedbiped: Don't know if I get you right, but to change the syntax of a wiki, all you have to do is to set the value of the 'syntax' key in g:vimwiki_list to 'markdown' or whatever.

blessedbiped:
Exactly my point. While I could understand that there is an option somewhere, the phrase, ```"set the value of the syntax key in g:vimwiki_list"``` means nothing to the average user (me). It took me blindly copying and pasting things from the help file into my vimrc to discover how this option is set.

a-b:
Is it possible to use ```vimwiki-todo-lists``` with markdown syntax?
[...]

PR #634, opened 2019-03-21
Adjust generated header behavior for TOC, links, tags, and diary (oh my!)

sstallion:
This PR corrects a number of annoyances I've encountered when using markdown and generated headers a la `VimiwikiTOC`, `VimwikiGenerateLinks`, `VimwikiGenerateTags`, and `VimwikiDiaryGenerateLinks`. The original behavior would place links on the line immediately after the header, which violates common markdown style guides. A new global variable was introduced named `markdown_header_style`, which indicates the number of newlines required after a header is generated. Both generated tags and links were updated to support behavior similar to `toc_header` and `toc_header_level`. Finally, some longstanding bugs that would emit unnecessary newlines after a generation before  content (if starting on the first line) and after (if starting after the first line) were fixed.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -9,6 +9,9 @@ endif
 let g:loaded_vimwiki_auto = 1
 
 
+let g:vimwiki_max_scan_for_caption = 5
+
+
 function! s:safesubstitute(text, search, replace, mode)
   " Substitute regexp but do not interpret replace
   let escaped = escape(a:replace, '\&')
@@ -351,23 +354,34 @@ function! vimwiki#base#generate_links()
 
   let bullet = repeat(' ', vimwiki#lst#get_list_margin()) . vimwiki#lst#default_symbol().' '
   for link in links
-    let abs_filepath = vimwiki#path#abs_path_of_link(link)
-    if !s:is_diary_file(abs_filepath)
+    let link_infos = vimwiki#base#resolve_link(link)
+    if !s:is_diary_file(link_infos.filename)
       if vimwiki#vars#get_wikilocal('syntax') == 'markdown'
         let link_tpl = vimwiki#vars#get_syntaxlocal('Weblink1Template')
       else
         let link_tpl = vimwiki#vars#get_global('WikiLinkTemplate1')
       endif
 
+      let link_caption = vimwiki#base#read_caption(link_infos.filename)
+      if link_caption == '' " default to link if caption not found
+        let link_caption = link
+      endif
+
       let entry = s:safesubstitute(link_tpl, '__LinkUrl__', link, '')
-      let entry = s:safesubstitute(entry, '__LinkDescription__', link, '')
+      let entry = s:safesubstitute(entry, '__LinkDescription__', link_caption, '')
       call add(lines, bullet. entry)
     endif
   endfor
 
-  let links_rx = '\m^\s*'.vimwiki#u#escape(vimwiki#lst#default_symbol()).' '
+  let links_rx = '\%(^\s*$\)\|\%('.vimwiki#vars#get_syntaxlocal('rxListBullet').'\)'
 
-  call vimwiki#base#update_listing_in_buffer(lines, 'Generated Links', links_rx, line('$')+1, 1, 1)
+  call vimwiki#base#update_listing_in_buffer(
+        \ lines,
+        \ vimwiki#vars#get_global('links_header'),
+        \ links_rx,
+        \ line('$')+1,
+        \ vimwiki#vars#get_global('links_header_level'),
+        \ 1)
 endfunction
 
 
@@ -1092,6 +1106,11 @@ function! vimwiki#base#update_listing_in_buffer(strings, start_header,
     let start_lnum = a:default_lnum
     let is_cursor_after_listing = ( cursor_line > a:default_lnum )
     let whitespaces_in_first_line = ''
+    " append newline if not replacing first line
+    if start_lnum > 1
+      keepjumps call append(start_lnum -1, '')
+      let start_lnum += 1
+    endif
   endif
 
   let start_of_listing = start_lnum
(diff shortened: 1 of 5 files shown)


sstallion:
Testing was done using each of the above reference commands, looking for incorrect formatting or style violations using markdown syntax. Behavior should be unchanged for other syntaxes.

ranebrown:
Ha! This is something I was planning on fixing myself. I'll take a look at your changes and start using it myself for a few days.

ranebrown:
@sstallion This is working well so far. What are your thoughts on modifying the generated links to not have an initial indent? For example, in the TOC level 1 headers would be flush with the `# Contents` and then lower level header indentation would be indented from there. The reason for this being if your settings use 4 spaces for an indent then many parsers consider this a code block ([link](https://commonmark.org/help/tutorial/09-code.html)). I think it would make sense to have that change as part of this PR.

Also, I know this is nit-picky but do you mind using more descriptive commit messages? Depending on how you are viewing the log 'Initial commit' isn't very helpful. Thanks!

sstallion:
@ranebrown Interesting. With the markdown viewers I've been using with `list_margin` set to `0` I haven't had any issues with rendering. I wonder if changing the defaults there would be enough? Sure thing - I'll amend the commit with a better message - sorry for that!

sstallion:
Here's an example using markdown with `toc_header_level` set to `2` and `list_margin` set to `0`:

    # Home

    ## Contents

    - [Home](#Home)
        - [Operating Systems](#Home#Operating Systems)
        - [Projects](#Home#Projects)
        - [Sundry](#Home#Sundry)

ranebrown:
Yep, I totally overlooked that option. Thanks for pointing it out, disregard my previous comment.

sstallion:
No worries! Given the amount of markdown-specific work that's been going on recently, I wonder if a mass refactoring of Vimwiki is overdue. I wonder how many people are using the other syntaxes these days? It would simplify things quite a bit if it were possible to settle on just one.

ranebrown:
I'm not opposed to the idea but I don't know how likely it is at this point. #15 has been around for a really long time...

sstallion:
Fair point - I definitely don't want to bandy the F-word about. It is getting tougher to make changes though and I think a lot of the logic could be condensed a bit to make supporting multiple syntax more seamless.

brennen:
> I wonder how many people are using the other syntaxes these days? It would simplify things quite a bit if it were possible to settle on just one.

As a point of anecdata, I'm using Vimwiki's default syntax for most of my work with the tool (and Markdown for a smaller secondary wiki that I publish as part of a website).  I wouldn't be completely opposed to switching to Markdown, but...

```
brennen@inertia 14:39:22 ~/notes/vimwiki (master 1/1 [ahead 2]) ▣ find . | wc -l
955
```

...those 955 pages would be a pretty substantial burden to transition to different markup without some kind of automatic conversion.  I have to imagine I'm not the only one in that situation.

I'll add that Vimwiki syntax has features I appreciate (TODO list partial completion states and such) that I'm not sure are mirrored effectively in the Markdown dialects I'm familiar with.

ranebrown:
@sstallion I haven't had as much time to look at this as I wanted but there does seem to be an issue with generating tags when using the default syntax.
```vim
Error detected while processing function vimwiki#tags#generate_tags[51]..vimwiki#vars
#get_syntaxlocal:
line   10:
E716: Key not present in Dictionary: WikiLinkTemplate1
E15: Invalid expression: g:vimwiki_syntax_variables[syntax][a:key]
E716: Key not present in Dictionary: WikiLinkTemplate1
E15: Invalid expression: g:vimwiki_syntax_variables[syntax][a:key]
```

Can you create a wiki using the default syntax and run through each command to make sure nothing else was broken?

sstallion:
Will do.

sstallion:
Looks like a typo; I called into `vimwiki#vars#get_syntax_local` rather than `vimwiki#vars#get_global` - the behavior between the original wiki syntax differs from other syntaxes it seems.

ranebrown:
Tags are working again for the default syntax.

l am noticing some interesting behavior when generating tags for a markdown page.

1. It looks like the first found header is used for the description `[description](link)`? That results in all the links appearing to be the same.
2. If a tag is placed directly after a header the link is made to the header instead of the tag. This seems to have been the case prior to your changes though. If you feel like taking a look at this that's great but otherwise I'll open a new issue.

```markdown
# Contents

- [Generated Sections Tests](#Generated Sections Tests)
- [Another Header](#Another Header)
    - [Chicken](#Another Header#Chicken)
    - [Beef](#Another Header#Beef)
        - [Steak](#Another Header#Beef#Steak)

# Generated Sections Tests

# Another Header

Various information about stuff.

## Chicken

- This is a list
- I like making lists
- Things should go here

## Beef

Beef comes from cows. :cow:

### Steak

Information about Steak.
Steak is a form of beef :forms:

# Generated Tags

## cow

- [Contents](index.md#Another Header#Beef) <<<< link to header not tag

## forms

- [Contents](index.md#forms) <<<< all descriptions use "Contents"
```

ranebrown:
Removed comment about tag linking to the header since that is the documented behavior if a tag is on the first two lines below a header.

sstallion:
I believe I fixed this behavior in #635 since it required some more significant changes to grab the correct caption. Could you give that a shot to see if it resolves your issue?

ranebrown:
Yep, that works for me too. Thanks

ranebrown:
Can you adjust the behavior of `g:vimwiki_tags_header_level`? If this is set to 6 it tries to generate the subheadings at level 7 and throws a bunch of errors. Maybe limit it to be 1-5 since there are subheadings?

sstallion:
Sure thing. Does this affect `diary_caption_level` as well? It looks like
it supports 6 levels too.

ranebrown:
No, that option just places links to all headers of the specified level in the diary index.

sstallion:
Gotcha. I'll push a change momentarily.

ranebrown:
Kinda just stumbled across this one. If you generate tags, generate links, then generate tags again the generated links are deleted. The old behavior added a newline and left the generated links. Are you seeing the same?

sstallion:
That's strange - this is working for me. Which syntax are you using?

sstallion:
I hate saying it, but it might be worth checking out #635 too. I stumbled across a mess of small bugs in how content is generated that I was able to clean up once I factored out content generation.

ranebrown:
Saw that behavior for both syntax variants. Seems to be related to having both the below options set:
```vim
let g:vimwiki_links_header_level = 6
let g:vimwiki_tags_header_level = 5
```

sstallion:
Ah, I think this is expected behavior. You didn't see it before because
links and tags were always at the same level. If you take a look at the
logic in `vimwiki#base#generate_links` and `vimwiki#tags#generate_tags` the
logic to replace content will grab all lines at the current header level
and below.

sstallion:
FWIW, this affects all generated content. You'd have a similar issue if the
TOC was set a level lower than links or tags and placed after those headers.

ranebrown:
Yep, I was just about to say the same. Not related to your changes just how generate tags works since it uses sub headings.

ranebrown:
Or manually created headers.

ranebrown:
Everything looks good. 🍻

PR #635, opened 2019-03-22
Support automatic generation of links and tags

sstallion:
This PR builds on #634 to add support for automatically generating links and tags, controlled by two new options: `auto_generate_links`, and `auto_generate_tags`, respectively. In order to accommodate  automatic behavior, `vimwiki#base#update_listing_in_buffer` was refactored to accept a funcref, which allows the caller to defer content generation. This allows generation to only update the buffer if a matching header is found, similar to `auto_toc`.

This was tested against a test markdown wiki with a dedicated `links.md` and `tags.md` file, which contain all generated links and tags. There is no noticeable change in performance with these options enabled for pages that do not contain generated headers.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -346,42 +346,47 @@ function! vimwiki#base#get_globlinks_escaped() abort
 endfunction
 
 
-function! vimwiki#base#generate_links()
-  let lines = []
+function! vimwiki#base#generate_links(create)
 
-  let links = vimwiki#base#get_wikilinks(vimwiki#vars#get_bufferlocal('wiki_nr'), 0)
-  call sort(links)
+  function! Generator() closure
+    let lines = []
 
-  let bullet = repeat(' ', vimwiki#lst#get_list_margin()) . vimwiki#lst#default_symbol().' '
-  for link in links
-    let link_infos = vimwiki#base#resolve_link(link)
-    if !s:is_diary_file(link_infos.filename)
-      if vimwiki#vars#get_wikilocal('syntax') == 'markdown'
-        let link_tpl = vimwiki#vars#get_syntaxlocal('Weblink1Template')
-      else
-        let link_tpl = vimwiki#vars#get_global('WikiLinkTemplate1')
+    let links = vimwiki#base#get_wikilinks(vimwiki#vars#get_bufferlocal('wiki_nr'), 0)
+    call sort(links)
+
+    let bullet = repeat(' ', vimwiki#lst#get_list_margin()) . vimwiki#lst#default_symbol().' '
+    for link in links
+      let link_infos = vimwiki#base#resolve_link(link)
+      if !s:is_diary_file(link_infos.filename)
+        if vimwiki#vars#get_wikilocal('syntax') == 'markdown'
+          let link_tpl = vimwiki#vars#get_syntaxlocal('Weblink1Template')
+        else
+          let link_tpl = vimwiki#vars#get_global('WikiLinkTemplate1')
+        endif
+
+        let link_caption = vimwiki#base#read_caption(link_infos.filename)
+        if link_caption == '' " default to link if caption not found
+          let link_caption = link
+        endif
+
+        let entry = s:safesubstitute(link_tpl, '__LinkUrl__', link, '')
+        let entry = s:safesubstitute(entry, '__LinkDescription__', link_caption, '')
+        call add(lines, bullet. entry)
       endif
+    endfor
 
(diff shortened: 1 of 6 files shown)


ranebrown:
@sstallion It looks like this isn't working for the default syntax. Can you test and make an update?

Also, the `auto_tags` option is a `BufWritePost` autocommand which causes the auto generated tags to not be up to date since that one is `BufWritePre`. I'm not seeing any reason why `auto_tags` couldn't be changed to a pre command as well.

sstallion:
Hi Rane - sure thing I'll double check the default syntax. I'll poke around
at auto_tags too - I agree, I don't think there's a reason why we couldn't
use `BufWritePre` there either.

sstallion:
@ranebrown This seems to be working for me. Are you sure you rebuilt your tags before generation? I took a look at `auto_tags` and I think `BufWritePre` is correct - `vimwiki#tags#update_tags()` works at a file level, so if the changes aren't written first, then they'll be missed when the tags database is updated.

ranebrown:
Here is an example of what I am seeing. With the below contents run `:VimwikiRebuildTags` and `VimwikiGenerateTags` and then save the file. Add another tag `:tag3:` and save. I think the expected behavior (if `auto_tags` and `auto_generate_tags` are both set) is to first rebuild the tags and then update the generated tags links. This isn't happening since `auto_tags` is a `BufWritePost` command. A second save will result in the new `:tag3:` being added to the list. Similiar behavor occurs if you delete `:tag3:` and save.

```
# Tags Test

Information about stuff.

Here is a :tag1:

Here is another :tag2:

This is a third tag that is going to be added.

# Generated Tags

## tag1

- [tag1](index.md#tag1)

## tag2

- [tag2](index.md#tag2)

```

I tested out the below change (`BufWritePre` instead of `BufWritePost`)  in
`ftplugin/vimwiki.vim` and the tag update worked as expected. Are you seeing
different behavior before or after making this change?

```vim
if vimwiki#vars#get_wikilocal('auto_tags')
  " Automatically update tags metadata on page write.
  augroup vimwiki
    au BufWritePre <buffer> call vimwiki#tags#update_tags(0, '')
  augroup END
endif
```

I don't think `getline` used by `vimwiki#tags#update_tags` requires the buffer to be written to disk first so a pre command should work no?

sstallion:
Ah! I completely missed the `getline` path - that makes a much more sense. I'll push a change to change `auto_tags` to `BufWritePre`.

ranebrown:
@sstallion Let me know when you get a chance to look at the default syntax issue and update this. Thanks!

sstallion:
Hi @ranebrown Things appear to be working for me. I created a test page with the exact contents you posted. After adding `:tag3:` and saving, generated tags are updated to contain a `tag3` section. Are you seeing something different with the last change I pushed?

ranebrown:
I was talking about the automatic tag and link generation not working for the default syntax. Did the change you make address that as well?

sstallion:
The following test is working for me:

    This is a test wiki.

    = This is a header =

    This is a :tag: and :another: tag.

    = Generated Links =
    - [[index]]

    = Generated Tags =
    == another ==
    - [[index#This is a header]]

    == tag ==
    - [[index#This is a header]]

ranebrown:
The automatic updating of links and tags works? I'm using these settings and it isn't working for the default syntax but works fine for markdown. What settings are you using?

```vim
let vimwiki_default = {}
let vimwiki_default.path = '~/Documents/test_wiki'
let vimwiki_default.path_html = '~/Documents/test_wiki/html'
let vimwiki.list_margin = 0
let vimwiki.auto_toc = 1
let vimwiki.auto_diary_index = 1
let vimwiki.auto_generate_links = 1
let vimwiki.auto_generate_tags = 1
let vimwiki.auto_tags = 1
let g:vimwiki_list = [vimwiki_default]
```

sstallion:
That's odd. Does it work if you use the test page I posted?

      let g:vimwiki_list = [
            \ {'path': '~/.vimwiki-test',
            \  'auto_diary_index': 1, 'auto_generate_links': 1,
            \  'auto_generate_tags': 1, 'auto_tags': 1, 'auto_toc': 1,
            \  'list_margin': 0,
            \ }]

ranebrown:
Strange, using your settings works.

sstallion:
Oh weird. What do you see if you `:echo g:vimwiki_list` using your original settings?

ranebrown:
Stupid mistake mixing `vimwiki` and `vimwiki_default` since I had two different wikis defined and copied the settings. Sorry for the confusion. It is working.

sstallion:
No worries! I'm glad you figured it out!

PR #641, opened 2019-03-27
Add an option to conceal preformatted text markers

gpanders:
This PR adds a feature allowing users to conceal code fence markers. 

For example, if you have the following in your wiki:

````
Some text

{{{python
def hello_world():
    print("Hello world!")
}}}
````

then setting `g:vimwiki_conceal_code_blocks` would conceal the `{{{python` and `}}}` lines to produce:
````
Some text


def hello_world():
    print("Hello world!")

````

This can make things look a lot nicer and more polished in wikis with many code blocks.

Note that this only applies to blocks that are syntax highlighted. So plain blocks that use just `{{{ }}}` (with no syntax highlighting) will not be concealed.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -994,11 +994,11 @@ function! vimwiki#base#nested_syntax(filetype, start, end, textSnipHl) abort
     let group='texMathZoneGroup'
   endif
 
-  let concealends = vimwiki#vars#get_global('conceal_code_blocks') ? ' concealends' : ''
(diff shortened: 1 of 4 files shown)


hq6:
Please fix the conflicts and force-push.

Also, what is the reasoning for handling syntax highlighted blocks differently than other pre-formatted text?

gpanders:
No particular reason, I can modify it to include all pre-formatted text.

gpanders:
Conflicts have been resolved. This is ready for merge.

hq6:
I have manually merged this in `e56c26c7ba892fc6ee073df0f2457d899d24d433` to have a cleaner history.

PR #646, opened 2019-03-31
Fix problems with commands not taking a count and mapping behavior.

### LINKED ISSUE #543: :VimwikiDiaryIndex not opening diary index for the "current" wiki.
The next quote is from the manual.

    :VimwikiDiaryIndex*
        Open diary index file of the current wiki.

So, according to this, it *should* open the diary index for the *current wiki* but it's not the case, it keeps opening the diary index for the *first wiki* in the list. *Just a remark, It doesn't say what should happen if you're not in a wiki, which I think, if it was working. should fallback to the first of the list.*

Although, it seems there is a mismatch with the section about the mapping to this function. For `<leader>wi` the manual has the following.

    [count]<Leader>wi or <Plug>VimwikiDiaryIndex
            Open diary index file of the [count]'s wiki.

            <Leader>wi opens diary index file of the first wiki from
            |g:vimwiki_list|.
            1<Leader>wi the same as above.
            2<Leader>wi opens diary index file of the second wiki from
            |g:vimwiki_list|.
            etc.
            To remap: >
            :nmap <Leader>i <Plug>VimwikiDiaryIndex

    See also |:VimwikiDiaryIndex|

So, according to the part `<Leader>wi` it should open the diary index for the *first wiki* from the list.

Now I'm ambivalent, according to the first one, it should be a bug. According to the later one, it should be a request. Indeed, I think opening the diary for the current wiki would be more convenient.

That said, I've noticed same dissonances in the documentation for the functions `:VimwikiMakeDiaryNote` and `:VimwikiTabMakeDiaryNote`. As for the `:VimwikiDiaryIndex` function, The function documentation sates that it will open a diary note for the *current wiki*. Again that's a mismatch against the documentation for the mapping to those functions which says the default (no wiki number) would open a note for the *first wiki*. **Unlike** `:VimwikiDiaryIndex`, it does open the diary for the *current wiki* for **both**, the mapping and the function. Honestly, I think it's nice that way but perhaps the manual needs some update.

Have a good day.

ranebrown:
This fixes a number of discrepancies in the documentation regarding the described behavior of commands and the actual behavior. For example, `:VimwikiIndex` was described as opening the index of the current wiki but it always opened the first wiki in the list. This also allows a count to open a specific wiki e.g `:VimwikiIndex 2`. This also affects a few diary related commands. Keymappings were also adjusted to have a more consistent behavior.

Fixes #543.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1267,17 +1267,21 @@ endfunction
 
 
 function! vimwiki#base#goto_index(wnum, ...)
+
+  " if wnum = 0 the current wiki is used
+  if a:wnum == 0
+    let idx = vimwiki#vars#get_bufferlocal('wiki_nr')
+    echom idx
+    if idx < 0  " not in a wiki
+      let idx = 0
+    endif
+  else
+    let idx = a:wnum - 1 " convert to 0 based counting
+  endif
+
   if a:wnum > vimwiki#vars#number_of_wikis()
     echomsg 'Vimwiki Error: Wiki '.a:wnum.' is not registered in your Vimwiki settings!'
     return
-  endif
-
-  " usually a:wnum is greater then 0 but with the following command it is == 0:
-  " vim -n -c ":VimwikiIndex"
-  if a:wnum > 0
-    let idx = a:wnum - 1
-  else
-    let idx = 0
   endif
 
   if a:0
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -286,8 +286,6 @@ function! vimwiki#diary#make_note(wnum, ...)
     return
   endif
 
-  " TODO: refactor it. base#goto_index uses the same
-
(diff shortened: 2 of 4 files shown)


hq6:
I manually merged this as `3396e87dbee617cbf0aa5b6ee9bcfe21270cc19a` after testing.

hq6:
Manually merged. @ranebrown , please let me know if you prefer that we stick to a consistent merge workflow instead of a mix of rebase and merge. I'm currently assuming our current mix is fine.

ranebrown:
@hq6 I think it makes more sense to merge pull requests instead of manually merging/rebasing them (unless there is a reason like lots of extra commits etc.). When reviewing issues/pull requests on GitHub it is much more obvious that something was implemented when it shows as merged.

hq6:
@ranebrown That's a fair point. I looked around but couldn't find a way to manually mark them as merged. It seems that this GitHub feature has been [sought after for 5 years](https://github.com/isaacs/github/issues/2) but never manifested.

I will create merge commits by default in the future.

PR #658, opened 2019-04-04
Fast aligns check

lyokha:
See description in the commit log.

To check this, create a very long table, say, with 1000 lines. Make sure that lines at the bottom of the table can be used for fast calculation of aligns, for example like in

```
|      uau | uauyaya         | ya      |
|       ua | uaua            | uaaua   |
| <cursor> |                 |         |
```
The line with `<cursor>` is empty, it only shows that cursor is inside it. The press `Tab` until another empty line appears. The motion should be very quick. As soon as you create another (the 3rd) empty line, the motion will get much slower because the fast check is now unable to figure out aligns by cells of the above 2 empty lines, and the algorithm turns to O(n) with traversal to the separator line of the table.

### DIFF
--- a/autoload/vimwiki/tbl.vim
+++ b/autoload/vimwiki/tbl.vim
@@ -254,13 +254,6 @@ function! s:get_rows(lnum, ...)
   while lnum <= line('$')
     let line = getline(lnum)
     if s:is_table(line)
-      if lnum == a:lnum && !s:is_separator(line)
-        let cells = vimwiki#tbl#get_cells(line)
-        let clen = len(cells)
-        let max_lens = repeat([0], clen)
-        let aligns = repeat(['left'], clen)
-        let line = s:fmt_row(cells, max_lens, aligns, 0, 0)
-      endif
       call add(rows, [lnum, line])
     else
       break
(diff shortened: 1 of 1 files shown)


hq6:
@lyokha For this and future PR's please clean up the git history a bit. You can do this setting your branch state to match vimwiki/dev **before adding new commits**, rather than merging vimwiki/dev. The commands to do this is the following, but just be careful because it can destroy local commits if you don't have them on a branch somewhere.
```
# Put your commits on a local branch somewhere
git checkout -b SavedCommits

# Switch to your dev branch
git checkout dev
git reset --hard upstream/dev

# Migrate the most recent commit over. Note that if there are multiple new commits, you may need to cherry-pick them in the order they were made.
git cherry-pick SavedCommits

# Push to your own remote
git push -f origin/dev
```

lyokha:
@hq6, thanks for the advice! I did exactly what you said, but still had to resolve conflicts anyway. This git thing is mysterious to me. Nevertheless, the diff looks good now.

Btw I am not sure about the name of the new function `get_cell_fast_aligns()`, perhaps `get_cell_aligns_fast()` is more correct.

lyokha:
Probably this happened because I did not fetch from upstream before doing all those git operations.

hq6:
The added code is very branch-heavy, so it might take some time to understand it and test it.

If others want to test it also and comment here, that would also be greatly appreciated.

ranebrown:
I gave this a quick test run and didn't see any issues.

Issue #456, opened 2018-03-16
Omnicomplete links issue (Markdown syntax on Windows)


RayDrew49:
Hi,

When trying to use omnicomplete to generate a list of links I'm getting the following:

[test omni](<c-x><c-o> gives a list of long paths like:

[test omni](../../../../../C:\Users\rdrew\onedrive\vimwiki\TaskPool

If I start the link with a hint like... [test omni](Tas<c-x><c-o> ... I get 'Pattern not found'

This is with markdown syntax on Windows 10. I get the same with or without the following in my vimrc:
set omnifunc=vimwiki#Complete

I'm new to both Vim and Vimwiki so hoping I'm missing something obvious.

Thanks for any help,

Ray

previsualconsent:
I'm getting the same thing.

ranebrown:
This is related to path handling which has issues in a few places on windows. See #478

ranebrown:
@RayDrew49 @previsualconsent Do you mind testing the changes I pushed to this [branch](https://github.com/vimwiki/vimwiki/tree/omni) and posting any results here or on PR #660? Thanks!

Issue #478, opened 2018-04-16
Refactor handling of paths and file names


EinfachToll:
A constant source of bugs is the sloppy, ad-hoc internal handling of file paths. Currently, paths are represented by strings and manipulated either "manually" using string operations or by using Vim's functions like fnamemodify(). Some edge cases which often go wrong now:
- path separators can be / or \ or both, depending on usage and platform
- what about \\/?
- a directory path can end in / or not
- unix paths can contain '~'
- Windows is normally case insensitive, Unix isn't
- etc.

We need a small path manipulating library which handles all the edge cases and hides the internals, so that the programmer isn't tempted to fuddle with strings.

tinmarino:
Dividing your notes:
1/ path separators can be `/` or `\` or both, depending on usage and platform
   - Windows still recognizes /. But windowsians may put their links with \ separators. There comes the problem
    - My solution would be  to unixify path before working internally in vimwiki#path and osxify its when giving it back.
        - Function definition : https://github.com/tinmarino/vimwiki/commit/81ab706e72c66300754899c54ad6536f3cbc2728
        - Example use : https://github.com/tinmarino/vimwiki/commit/6d1d4d20b05e5dda5ecc2fac156f27379431672d 

2/ what about `\/`? `\' in linux is a legal character, shoudl it be allowed or replaced ?
    - In my opinion: `\/` -> `//` -> `/` on linux (so not allowing `\` in dirnames)
                                  -> `\\` -> `\` on window because it is the windows default path normalization

3/ a directory path can end in `/` or not 

4/ unix paths can contain `~`
   - `let path = expand(path)` solves that, for better architecture, this function must be used ONLY in `vimwiki#path` autoload
 
5/ Windows is normally case insensitive, Unix isn't
    - Already solved by `function! vimwiki#path#is_equal(p1, p2)`

levirs565:
I have been refractor handling subdir in  [my pull request](https://github.com/vimwiki/vimwiki/pull/1057).

tinmarino:
This may be the most important issue of vimwiki.
I propose that the

1. output of the generated path are all with `/` and the
2. parsing considers igually `/` and `\` with some `[/\\]`


Image a path with `a\b/c` and c renamed to `../e/f` -> `a\e/f` and this mixing is hard to track.

PR #660, opened 2019-04-07
Omnicompletion fix for Windows

ranebrown:
This is a temporary fix for #456. I tested the changes on a Windows VM and omnicompletion seems to be working for markdown and default syntax. I did not encounter any regressions on a non-Windows environment but it could use some more testing.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -486,6 +486,10 @@ function! vimwiki#base#get_wikilinks(wiki_nr, also_absolute_links)
   let result = []
   for wikifile in files
     let wikifile = fnamemodify(wikifile, ':r') " strip extension
+    if vimwiki#u#is_windows()
+      " TODO temporary fix see #478
+      let wikifile = substitute(wikifile , '/', '\', 'g')
+    endif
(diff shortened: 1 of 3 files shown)


Nudin:
Do we maybe want to write a utility function  to handle path splitting and joining instead of having the case distinction everywhere?

hq6:
That is likely a good idea.

ranebrown:
I could create a function but is it really worth it since this should be deleted with #478? I don't know how much progress @EinfachToll has made on that front put there is a [branch](https://github.com/vimwiki/vimwiki/tree/path-handling) where at least some work has been done. I don't really want to spend more time working on something that is temporary and someone else is trying to fix in another way. I guess it is also somewhat dependent on the timeframe for #478 and how many similar issue crop up.

What's the consensus?

hq6:
I am fine with putting this in as is with the understanding that we hope to clean it up eventually anyways.
@Nudin, any thoughts?

ranebrown:
@Nudin just wanted to ping you again on this.

Nudin:
Since we don't know the status of #478 I'm okay with using this as a temporary solution, even through it's kinda ugly.

Issue #245, opened 2016-09-28
Automatically Title New Wiki Pages


KTSCode:
I think it would be awesome to automatically title each new page that is created from a wiki link. 

I just switched from xolox/vim-notes to vimwiki, and I really miss this feature, it helped a lot with locational awareness. Since switching to vimwiki, I've found myself manually titling every new page I create. 

It would be extra useful if the title was highlighted, and was used as the page title when converted to html.

EinfachToll:
What do you mean by title? Obviously, the link name becomes the file name of the newly created page. What else do you want?

bhrutledge:
I think the idea is that pressing Enter on `[[Title]]` would create `Title.wiki`, with the first line being `= Title =`. Similarly, it would be nice if `:VimwikiMakeDiaryNote` would add something like `= 2016-11-29 =` to new notes; it would be especially awesome if the format could be configurable.

On a related note, it would be cool if `:VimwikiGenerateLinks` used the first header text as the link description, similar to `:VimwikiDiaryGenerateLinks`.

bhrutledge:
I'm guessing this could be accomplished independently of vimwiki, using [`:autocmd`](http://stackoverflow.com/questions/162617/how-can-i-automatically-add-some-skeleton-code-when-creating-a-new-file-with-vim) or something like [vim-skeleton](https://github.com/noahfrederick/vim-skeleton).

aq2:
old thread, i know, but a very simple two liner solution is
```
autocmd BufNewFile *.wiki :r! echo = %:t:r =
autocmd BufNewFile *.wiki :norm kddo
```
i am no vim scripter, so i am sure there are much better ways of doing this, but the first line automatically inserts the title (wrapped with '=') into the document, but for some reason inserts a new line before it. So the second autocommand deletes it - again i am sure there are better ways - all comments and suggestions welcome!

ranebrown:
@KTSCode @bhrutledge @aq2 I would appreciate it if any of you could test the implementation in PR #661 and post any results here or in the PR. Thanks!

aq2:
well that was simple!

seems to work fine here - i made a comment on the PR

aq2:
[diary-ex](https://user-images.githubusercontent.com/13156649/56463407-a50b4180-63cb-11e9-9782-7931c313ba95.png)

PR #661, opened 2019-04-09
Automatically generate a header for new wiki pages

ranebrown:
Add a new option `g:vimwiki_auto_header` that will automatically generate a level 1 header using the filename when creating a new file. The behavior is disabled for the wiki index and diary index. The function takes into account the options `vimwiki-option-links_space_char` and `g:vimwiki_markdown_header_style` when generating the header.

Tested with default and markdown syntax on macOS and Windows.

Related issue #245.

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -317,11 +317,11 @@ NORMAL MODE                                           *vimwiki-local-mappings*
                         To remap: >
                         :nmap <Leader>wp <Plug>VimwikiPrevLink
 <
-gnt			*vimwiki_gnt*
-			Find next unfinished task in the current page.
-			Maps to |:VimwikiNextTask|
-			To remap: >
-			:nmap <Leader>nt <Plug>VimwikiNextTask
+gnt                     *vimwiki_gnt*
+                        Find next unfinished task in the current page.
+                        Maps to |:VimwikiNextTask|
+                        To remap: >
+                        :nmap <Leader>nt <Plug>VimwikiNextTask
 <
                         *vimwiki_<Leader>wd*
 <Leader>wd              Delete wiki page you are in.
@@ -3186,6 +3186,22 @@ values are from 0 to 2.
 The default is 1.
 
 
+------------------------------------------------------------------------------
+*g:vimwiki_auto_header*
+
+Set this option to 1 to automatically generate a level 1 header when creating
+a new wiki page. This option is disabled for the wiki index and the diary
+index. Spaces replaced with |vimwiki-option-links_space_char| are reverted
+back to spaces in the generated header, which will match the filename
+except for the characters that were reverted to spaces.
+
+For example, with `links_space_char` set to `'_'` creating a link from the text
+`foo bar link` would result in `[[foo_bar_link]]` and the file
+`foo_bar_link.wiki`. The generated header would be `= foo bar link =`
+
+The default is 0.
+
+
 ==============================================================================
 13. Getting help                                                *vimwiki-help*
 
@@ -3281,6 +3297,8 @@ https://github.com/vimwiki-backup/vimwiki/issues.
 2.5 (in progress)~
 
 New:~
+    * PR #661: Add option |g:vimwiki_auto_header| to automatically generate
+      a level 1 header for new wiki pages.
     * PR #665: Integration with vimwiki_markdown gem
       https://github.com/patrickdavey/vimwiki_markdown
       This provides the |vimwiki-option-html_filename_parameterization|
--- a/autoload/vimwiki/vars.vim
+++ b/autoload/vimwiki/vars.vim
@@ -143,6 +143,7 @@ function! s:read_global_settings_from_user()
   let global_settings = {
         \ 'CJK_length': {'type': type(0), 'default': 0, 'min': 0, 'max': 1},
         \ 'auto_chdir': {'type': type(0), 'default': 0, 'min': 0, 'max': 1},
+        \ 'auto_header': {'type': type(0), 'default': 0, 'min': 0, 'max': 1},
         \ 'autowriteall': {'type': type(0), 'default': 1, 'min': 0, 'max': 1},
         \ 'conceallevel': {'type': type(0), 'default': 2, 'min': 0, 'max': 3},
         \ 'conceal_onechar_markers': {'type': type(0), 'default': 1, 'min': 0, 'max': 1},
--- a/plugin/vimwiki.vim
+++ b/plugin/vimwiki.vim
@@ -239,6 +239,50 @@ if !exists("*VimwikiWikiIncludeHandler")
 endif
 
 
+" write a level 1 header to new wiki files
+" a:fname should be an absolute filepath
+function! s:create_h1(fname)
+  if vimwiki#vars#get_global('auto_header')
+    let idx = vimwiki#vars#get_bufferlocal('wiki_nr')
+
+    " don't do anything for unregistered wikis
+    if idx == -1
+      return
+    endif
+
+    " don't create header for the diary index page
+    if vimwiki#path#is_equal(a:fname,
+          \ vimwiki#vars#get_wikilocal('path', idx).vimwiki#vars#get_wikilocal('diary_rel_path', idx).
+          \ vimwiki#vars#get_wikilocal('diary_index', idx).vimwiki#vars#get_wikilocal('ext', idx))
+      return
+    endif
+
+    " get tail of filename without extension
+    let title = expand('%:t:r')
+
+    " don't insert header for index page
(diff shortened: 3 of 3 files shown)


aq2:
seems to work fine with new links, including those with space characters, and diary date entries.

tested with vimiwiki syntax on linux

anything else you particularly want me to test?

ranebrown:
@aq2 if you feel like it you could try the different syntax options, creating files in nested directories, files with strange characters in the name etc. Also just using the option in your normal workflow for a couple of days. Thanks for giving it a try so far!

doc/vimwiki.txt
```diff
@@ -3150,6 +3150,17 @@ values are from 0 to 2.
 The default is 1.
 
 
+------------------------------------------------------------------------------
+*g:vimwiki_auto_header*
+
+Set this option to 1 to automatically generate a level 1 header when creating
+a new wiki page. This option is disabled for the wiki index and the diary
+index. Spaces replaced with |vimwiki-option-links_space_char| are reverted.
```
hq6: Why is it a good idea to revert spaces here, rather than still respecting the value of `vimwiki-option-links_space_char`?

autoload/vimwiki/base.vim
```diff
@@ -770,6 +770,47 @@ function! vimwiki#base#check_links()
   endif
 endfunction
 
+" write a level 1 header to new wiki files
+function! s:create_h1(fname, create)
+  if vimwiki#vars#get_global('auto_header')
+    if a:create
```
hq6: Nit: Extra layer of nesting seems unnecessary.

autoload/vimwiki/base.vim
```diff
@@ -770,6 +770,47 @@ function! vimwiki#base#check_links()
...
+      " don't insert header for index page
+      if title ==# vimwiki#vars#get_wikilocal('index', idx)
+        return
+      endif
+
+      " don't substitute space char for diary pages
+      if title !~# '^\d\{4}-\d\d-\d\d'
+        " NOTE: it is possible this could remove desired characters if the 'links_space_char'
+        " character matches characters that are intentionally used in the title.
+        let title = substitute(title, vimwiki#vars#get_wikilocal('links_space_char'), ' ', 'g')
```
hq6: This seems undesirable, since I think the `vimwiki-option-links_space_char` was intended to be a one-way transformation because of precisely the problem you pointed out in the Note.

doc/vimwiki.txt
ranebrown: I guess it's more of a personal preference for readability of the header. I like to have my filenames as `My_File_Name.md` but the header looks better if spaces are actually used.

```markdown
# My File Name <-- desired

# My_File_Name
```

The case I ran into was with the diary pages (they are now specifically excluded though). In that case the filename is something like `2019-04-19.md` and if `links_space_char` is set to `-` the headers ended up like this:

```markdown
# 2019 04 19

# 2019-04-19 <-- desired
```

I think the probability of inadvertently changing characters is fairly low and the behavior is documented. Even if it does happen the usability isn't affected since it is just the appearance of a header title.

autoload/vimwiki/base.vim
ranebrown: Good point. I'll change it to `vimwiki#vars#get_global('auto_header') && a:create`

autoload/vimwiki/base.vim
ranebrown: See my comment about the description in the help file.

doc/vimwiki.txt
hq6: I just tested and found the following documentation a little misleading.
```
Spaces replaced with vimwiki-option-links_space_char are reverted.
The title will match the filename.
```

Consider a `.vimrc` with the following options.
```
let g:vimwiki_links_space_char = '_'
let g:vimwiki_auto_header = 1
```

Start with the following text, visually select it and turn it into a link.

```
foo bar blah blah
``` 

At this point, my text looks like this:
```
[[foo_bar_blah_blah]]
```

If I type `Enter`, then I get a page with a header that looks like this.
```
= foo bar blah blah =
```

Now we have a file and a link with the name `foo_bar_blah_blah.wiki` and a first level header that has `foo bar blah blah`.  This behavior is what you described, but it does not match the documentation:
 * The documentation says "Spaces replaced with vimwiki-option-links_space_char are reverted.", but this is true only of the header, not of the filename. I propose appending "in the header" and possibly adding the caveat about changing intentional characters into spaces into the documentation.
 * The documentation says "The title will match the filename." but this is not true because the filename will contain `_`'s and the title contains spaces.

I legitimately could not tell what the behavior was going to be from the documentation, which is why I had to experiment to find out, so perhaps an illustrative example in the documentation would make things more clear?

---

There's another unrelated behavior I observed and found a little strange, but it may be by design:

 * If I create a file by following a link, the header is generated. 
 * However, if I create a file by typing `vim X.wiki`, where `X.wiki` does not previously exist, then no header is generated.

Is this difference by design or oversight?

hq6:
I just tested and found the following documentation a little misleading.
```
Spaces replaced with vimwiki-option-links_space_char are reverted.
The title will match the filename.
```

Consider a `.vimrc` with the following options.
```
let g:vimwiki_links_space_char = '_'
let g:vimwiki_auto_header = 1
```

Start with the following text, visually select it and turn it into a link.

```
foo bar blah blah
``` 

At this point, my text looks like this:
```
[[foo_bar_blah_blah]]
```

If I type `Enter`, then I get a page with a header that looks like this.
```
= foo bar blah blah =
```

Now we have a file and a link with the name `foo_bar_blah_blah.wiki` and a first level header that has `foo bar blah blah`.  This behavior is what you described, but it does not match the documentation:
 * The documentation says "Spaces replaced with vimwiki-option-links_space_char are reverted.", but this is true only of the header, not of the filename. I propose appending "in the header" and possibly adding the caveat about changing intentional characters into spaces into the documentation.
 * The documentation says "The title will match the filename." but this is not true because the filename will contain `_`'s and the title contains spaces.

I legitimately could not tell what the behavior was going to be from the documentation, which is why I had to experiment to find out, so perhaps an illustrative example in the documentation would make things more clear?

---

There's another unrelated behavior I observed and found a little strange, but it may be by design:

 * If I create a file by following a link, the header is generated. 
 * However, if I create a file by typing `vim X.wiki`, where `X.wiki` does not previously exist, then no header is generated.

Is this difference by design or oversight?

ranebrown:
@hq6 What do you think about changing the help file to this:

```help
*g:vimwiki_auto_header*

Set this option to 1 to automatically generate a level 1 header when creating
a new wiki page. This option is disabled for the wiki index and the diary
index. Spaces replaced with |vimwiki-option-links_space_char| are reverted.
The title will match the filename except for the characters that were reverted
to spaces.

For example, with `links_space_char` set to `'_'` creating a link from the text
`foo bar link` would result in `[[foo_bar_link]]` and the file
`foo_bar_link.wiki`. The generated header would be `= foo bar link =`

The default is 0.
```

ranebrown:
> If I create a file by following a link, the header is generated.
However, if I create a file by typing vim X.wiki, where X.wiki does not previously exist, then no header is generated.
Is this difference by design or oversight?

That's an oversight. I didn't consider that files created that way or using `:e file.wiki` would need handled differently since I never create files that way. Good catch! I'll take a look into how to handle those cases.

hq6:
I think that is much more clear, and it would be even more clear if you added the words "back to spaces in the generated header." after " are reverted."

ranebrown:
@hq6 I think I covered the additional cases. I tested out the following:

1. Opening a file in a registered wiki with `vim file.md`
2. Opening a file in a registered wiki with `:e file.md`
3. Creating files via a link.
4. Opening a file in a temporary wiki.
5. Setting `let g:vimwiki_global_ext = 0` and verifying the same file as opened in (4) doesn't get a header created.

Let me know if you run into any problems or see something else that is missing. Thanks for testing!

@aq2 If you feel like testing with these new changes that would be helpful as well.

aq2:
I tried:
1 - creating files with .wiki extension from linux command line, with and without space_char
2 - creating links from wiki pages with various space_chars, nested folders and ../ links
3 - creating files using e: foo/bar/baz-baz.wiki
4 - as above with multiple space_chars - eg foo____bar.wiki

Everything seemed to work OK. :)

I still think the help text is a bit confusing - one sentence refers to the 'generated header', and the next mentions the 'title', and it was unclear to me that these are the same thing.

```
Spaces replaced with |vimwiki-option-links_space_char| are reverted
back to spaces in the generated header, which will match the filename
except for the characters that were reverted to spaces.
```

hq6:
@ranebrown This appears to work with the minimum vimrc setup, but opening a file with `vim foo.wiki` does not generate a header when I use it with my personal vimrc. I am still diagnosing the reason.

PR #663, opened 2019-04-09
Add option to not conceal one-character markers

### LINKED ISSUE #315: Don't conceal one-character markers
Add an option to disable conceiling one-char markers? I want typefaces to not be conceiled but URLs still shortened.

padowi:
Adds new configuration variable, "conceal_onechar_markers", defaulting
to on (preserving default behaviour)

Adds if-statement around relevant parts of code (as suggested in the
issue), which uses the new configuration variable.

Fix #315 - Don't conceal one-character markers

Testing done:

I have checked that default behaviour is preserved, both in the case where the configuration variable is present in .vimrc (with default value) as well as when it is not present.

I have also checked that the intended behaviour is exhibited when setting the value to 0, regardless of what setting vimwiki_conceallevel is running with (0..3).

Tested on Vim 8.1

### DIFF
--- a/syntax/vimwiki.vim
+++ b/syntax/vimwiki.vim
@@ -155,24 +155,26 @@ endfor
 " possibly concealed chars
 let s:conceal = exists("+conceallevel") ? ' conceal' : ''
 
-execute 'syn match VimwikiEqInChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_eqin').'/'.s:conceal
-execute 'syn match VimwikiBoldChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_bold').'/'.s:conceal
-execute 'syn match VimwikiItalicChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_italic').'/'.s:conceal
-execute 'syn match VimwikiBoldItalicChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_bolditalic').'/'.s:conceal
-execute 'syn match VimwikiItalicBoldChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_italicbold').'/'.s:conceal
-execute 'syn match VimwikiCodeChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_code').'/'.s:conceal
-execute 'syn match VimwikiDelTextChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_deltext').'/'.s:conceal
-execute 'syn match VimwikiSuperScript contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_superscript').'/'.s:conceal
-execute 'syn match VimwikiSubScript contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_subscript').'/'.s:conceal
+if vimwiki#vars#get_global('conceal_onechar_markers') == 1
+  execute 'syn match VimwikiEqInChar contained /'.
+        \ vimwiki#vars#get_syntaxlocal('char_eqin').'/'.s:conceal
+  execute 'syn match VimwikiBoldChar contained /'.
+        \ vimwiki#vars#get_syntaxlocal('char_bold').'/'.s:conceal
+  execute 'syn match VimwikiItalicChar contained /'.
+        \ vimwiki#vars#get_syntaxlocal('char_italic').'/'.s:conceal
+  execute 'syn match VimwikiBoldItalicChar contained /'.
+        \ vimwiki#vars#get_syntaxlocal('char_bolditalic').'/'.s:conceal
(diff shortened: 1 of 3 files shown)


syntax/vimwiki.vim
```diff
@@ -155,24 +155,26 @@ endfor
...
-      \ vimwiki#vars#get_syntaxlocal('char_italicbold').'/'.s:conceal
-execute 'syn match VimwikiCodeChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_code').'/'.s:conceal
-execute 'syn match VimwikiDelTextChar contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_deltext').'/'.s:conceal
-execute 'syn match VimwikiSuperScript contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_superscript').'/'.s:conceal
-execute 'syn match VimwikiSubScript contained /'.
-      \ vimwiki#vars#get_syntaxlocal('char_subscript').'/'.s:conceal
+if vimwiki#vars#get_global('conceal_onechar_markers') == 1
```
ranebrown: Not a big deal but you don't need the `== 1` part of this.

ranebrown:
@padowi Can you update the changelog in `doc/vimwiki.txt` and add your name to the contributors section (if you want).

Did you only test with the default syntax or did you also try out the other options?

padowi:
@ranebrown I am not sure I understand the question about default syntax. I may be too much of a noob here. In my .vimrc I tried all of the combinations for vimwiki_conceallevel (0..3) and conceal_onechar_markers (0..1)

vimwiki_conceallevel functionality appears unaltered (as compared to earlier version of vimwiki without my modifications at all) as long as conceal_onechar_markers use the default value (1). 

Setting conceal_onechar_markers does have effect on vimwiki_conceallevel (in that one character markers, if conceallevel would have hidden them, are now not concealed). Since this was the whole point, it ought be considered OK ;)

ranebrown:
> I am not sure I understand the question about default syntax.

No worries. Vimwiki supports multiple syntax options. It is a good idea to make sure changes work under all variants. Take a look at the documentation [here](https://github.com/vimwiki/vimwiki/blob/dev/doc/vimwiki.txt#L2045-L2057)

Also the changelog update should be made [here](https://github.com/vimwiki/vimwiki/blob/dev/doc/vimwiki.txt#L3246). Let me know if you have any questions.

padowi:
Aha! I see :)

It is getting a bit too late in the evening here, so will have to continue this tomorrow, but then I will run more tests (different syntaxes), and also update changelog accordingly.

Thank you for your time so far :)

padowi:
Ok, changelog added, and more tests performed.

The markdown syntax behaves just like default vimwiki syntax does, but there is one deviation in the mediawiki syntax.

According to https://www.mediawiki.org/wiki/Help:Formatting#Text_formatting_markup, it should be possible to create italicized bold text using four single-quotes, e.g. ''''lala''''.

Now, it could be argued that this is in fact not a single-character marker, but I guess that it still goes in the spirit of a single-character marker.

I don't believe this deviation to be completely related to this PR, because even with the default value (conceal) a singular single-quote is displayed on both sides of the marked up text. It appears that vimwiki by default does not consider four single-quotes to be valid markup.

e.g.:
If I write ''''lala'''', and have configured to conceal markers, it will still show 'lala', indicating to me that vimwiki recognizes three single-quotes as a marker (which I believe is bold/strong in mediawiki markup).

If I configure to show single-character markers, it will show all of them (''''lala''''), as intended/expected


Either way, the addition of vimwiki_conceal_onechar_markers makes it possible to show the remaining three (concealed) single-quotes, and since the fourth isn't concealed by default, the result is that with this addition, vimwiki will correctly show all four quotes if we opt to not conceal the single-character markers.

So the end result seems to be wholly as intended, but there may be a bug in how mediawiki syntax is handled (depending on if em+strong via '''' should be handled or not)

Nudin:
> According to https://www.mediawiki.org/wiki/Help:Formatting#Text_formatting_markup, it should be possible to create italicized bold text using four single-quotes, e.g. ''''lala''''.
Correction: It's five not four single-quotes.

padowi:
> 
> 
> > According to https://www.mediawiki.org/wiki/Help:Formatting#Text_formatting_markup, it should be possible to create italicized bold text using four single-quotes, e.g. ''''lala''''.
> 
> Correction: It's five not four single-quotes.

Gah... ok so TIL I can't count past 3 or 4... ;D
Will test some more , thanks for the patience ;D

padowi:
And having used 5 single-quotes instead of 4, I feel pretty satisfied that it works for mediawiki syntax as well :)

ranebrown:
@padowi Thanks for the contribution!

padowi:
> @padowi Thanks for the contribution!

My pleasure :) thanks for the support and sorry for this late reply, life, and a flu, came in the way ;D

PR #665, opened 2019-04-09
Add html_filename_parameterization option.

### LINKED ISSUE #129: question: wikiname -> html links converstion
Hi there,

I love vimwiki and use it all the time.  I currently use it in (github) markdown format and use a little rubygem I created to convert the `md` to `html` files.  That's all working fine.  As a part of the linking process, I take the `title` of the page, and use [parameterize](http://apidock.com/rails/ActiveSupport/Inflector/parameterize) to convert the names to be web safe.  So, for example `[[My Happy Page]]` becomes - `my-happy-page.html`

My ruby script runs quite slowly however, but most of that is down to the fact that `vimwiki` doesn't correctly skip the files, as, of course, the names don't match.  I was wondering if you would accept a pull request to at least make the sanitization configurable.  Say it could have a default, but if you supplied a function name it would call out to that.

My own first very crude approximation is here:
https://github.com/patrickdavey/vimwiki-1/blob/with_my_sanitized/autoload/vimwiki/html.vim#L242-248

With that in place it drops from a minute to a few seconds to generate changed files. I need to work out how to correctly delete files, that's not done at the moment, but it'll use similar code I guess.

Anyway, love vimwiki, happy to maintain my own fork, but thought I'd ask.

patrickdavey:
This adds the vimwiki_local `html_filename_parameterization` option (0|1)
which performs the same sanitization of filenames as does the vimwiki_markdown gem.

For example, if your file is called "My File.md" it will be written out
as "my-file.html".

If the html_filename_parameterization options is enabled, we also _do
not delete_ non-matching html files when the VimwikiAllToHtml command is run.

Closes #129

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -239,8 +239,15 @@ function! s:is_html_uptodate(wikifile)
   endif
 
   let wikifile = fnamemodify(a:wikifile, ":p")
-  let htmlfile = expand(vimwiki#vars#get_wikilocal('path_html') .
-        \ vimwiki#vars#get_bufferlocal('subdir') . fnamemodify(wikifile, ":t:r").".html")
+
+  if vimwiki#vars#get_wikilocal('html_filename_parameterization')
+    let parameterized_wikiname = s:parameterized_wikiname(wikifile)
+    let htmlfile = expand(vimwiki#vars#get_wikilocal('path_html') .
+          \ vimwiki#vars#get_bufferlocal('subdir') . parameterized_wikiname)
+  else
+    let htmlfile = expand(vimwiki#vars#get_wikilocal('path_html') .
+          \ vimwiki#vars#get_bufferlocal('subdir') . fnamemodify(wikifile, ":t:r").".html")
+  endif
 
   if getftime(wikifile) <= getftime(htmlfile) && tpl_time <= getftime(htmlfile)
     return 1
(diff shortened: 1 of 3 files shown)


ranebrown:
@patrickdavey Overall this looks pretty good. The only changes I see off the top of my head are to update the [changelog](https://github.com/vimwiki/vimwiki/blob/82ba1e99bf5deb963d93db781b494a7d25dfd690/doc/vimwiki.txt#L3247) and add a link to your `vimwiki_markdown` project.

It might make sense to add a short section like done [here](https://github.com/vimwiki/vimwiki/blob/82ba1e99bf5deb963d93db781b494a7d25dfd690/doc/vimwiki.txt#L1794-L1802) with a brief overview of your project and the link.

It will probably be a few days before I get to any testing with this.

patrickdavey:
Updated the docs with the link to project + changelog section.

Nudin:
Seems good. The only thing I wonder about: I guess that sooner or later someone will asks for an option to convert spaces to underscore instead of to a dash. What's our plan for that? Use '2' for this variant?

ranebrown:
@Nudin This is for integration with [vimwiki_markdown](https://github.com/patrickdavey/vimwiki_markdown) to convert markdown syntax files to html. It doesn't really have a use outside of that currently so I don't think that we need to worry about adding another variant.

ranebrown:
@patrickdavey I tested this and I don't see any issues. I think it would make sense to slightly modify the help description of the documentation to say it prevents the html files from being regenerated every time conversion is done. After that I think its good to merge!

patrickdavey:
@ranebrown I've altered the text slightly - see https://github.com/vimwiki/vimwiki/blob/5f01182468479a36a82def124f1b597945434924/doc/vimwiki.txt#L2445-L2449

If you want it changed again, can you just let me know what you'd like and I'll put it in there (or just merge it and then change it to what you like)

PR #672, opened 2019-04-15
Ignore verbatim blocks when scanning tags

sstallion:
This PR addresses a long-standing TODO when in `s:scan_tags` to skip over verbatim blocks when scanning for tags. Support for inline verbatim blocks is also supported for markdown wikis.

### DIFF
--- a/autoload/vimwiki/tags.vim
+++ b/autoload/vimwiki/tags.vim
@@ -85,8 +85,27 @@ function! s:scan_tags(lines, page_name)
   let PROXIMITY_LINES_NR = 2
   let header_line_nr = - (2 * PROXIMITY_LINES_NR)
 
+  let verbatim_block = 0
+
   for line_nr in range(1, len(a:lines))
     let line = a:lines[line_nr - 1]
+
+    " ignore verbatim blocks
+    if verbatim_block
+      if match(line, vimwiki#vars#get_syntaxlocal('rxPreEnd')) != -1
+        let verbatim_block = 0
+      endif
+      continue
+    else
+      if match(line, vimwiki#vars#get_syntaxlocal('rxPreStart')) != -1
+        let verbatim_block = 1
+        continue
+      endif
+      if vimwiki#vars#get_wikilocal('syntax') == 'markdown' &&
+            \ match(line, vimwiki#vars#get_syntaxlocal('rxPreInline')) != -1
+        continue
+      endif
+    endif
 
     " process headers
     let h_match = matchlist(line, rxheader)
@@ -111,8 +130,6 @@ function! s:scan_tags(lines, page_name)
       endif
       continue " tags are not allowed in headers
     endif
-
-    " TODO ignore verbatim blocks
 
     " Scan line for tags.  There can be many of them.
     let str = line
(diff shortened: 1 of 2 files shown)


ranebrown:
The only problem I see is that tags are still highlighted in blocks indented by 4 spaces. That could be confusing since the tag appears to be valid but then doesn't show up in the generated tags list. The syntax probably needs modified to highlight [indented code blocks](https://spec.commonmark.org/0.12/#indented-code-block) (block of 1 or more lines indented by four _or more_ spaces and surrounded by blank lines) the same as fenced code blocks using backticks \`. It would make sense to add support for [fenced code blocks](https://spec.commonmark.org/0.12/#fenced-code-blocks) with 3 tildes `~` as well.

Thoughts?

sstallion:
I stopped short of fixing this - I think there's another issue we'd need to
sort out as well. List items can be indented by four or more spaces as well
(not to mention list characters can be escaped if needed). This seems to be
getting awfully close to seeking perfection for markdown syntax, which
sounds like a rather unpleasant undertaking in vimscript. Do you have a
feeling for where the line should be drawn?

ranebrown:
I don't have a great answer. I would like to get as many things as close to the spec as possible but it does get pretty hairy quickly (there are 17 rules and about a million examples for emphasis).

At a minimum I think the tags shouldn't be highlighted in indented code block (four spaces) case if they are ignored in the generated output. If you think that is too difficult at this point it might make more sense to allow tags in that case and only exclude the fenced code blocks which is much easier to handle.

ranebrown:
Also, before I forgot - can you update the changelog with a quick note when you get a chance?

sstallion:
Sure thing. I'll get indenting working as well. That shouldn't be too bad.

sstallion:
@ranebrown I took a look this evening - this might take more surgery than I have time to do. We could update `rxTags` to handle indented code blocks, but dealing with fenced code blocks will need a bit more work. Would it be reasonable to merge this PR as-is and spin off an issue to updating the highlighting rules for tags?

ranebrown:
I thought the fenced code blocks already ignored the tag syntax highlighting? Only the indented code block exclusion needed added. Am I missing something?

sstallion:
Sorry for taking so long to get back to this. I think the issue is that we can't really negate indented code blocks without impacting other use cases for indentation (i.e. indented lists). If we add a syntax region for rxPreInline it will also match against anything that is indented (see attachment).
Screenshot from 2019-04-27 11-29-03

ranebrown:
Shouldn't the requirement that the indented code block has a leading and trailing blank line fix the conflicts with lists?

ranebrown:
I think this would work:

```vim
syntax match VimwikiMdIndentedCode '^\s*\n\(\(\s\{4,}[^ ]\|\t\+[^\t]\).*\n\)\+\(^\s*\n\)'
hi def link VimwikiMdIndentedCode PreProc
```

ranebrown:
@sstallion Have you had time to take a look at this?

sstallion:
I should be able to take a look this weekend - sorry for the delay, work
has been picking up a bit lately.

ranebrown:
No worries, just wanted to check back on it.

ranebrown:
@sstallion I made some modifications and merged this if you want to test it out.

Issue #477, opened 2018-04-15
Use name of wiki for interwiki links


davidlmontgomery:
TLDR: Enhancement proposal: allow wikis to have names, and support using these names in links.

I have found vimwiki's interwiki link schema ("wiki" + number of wiki +  ":") to be inconvenient. For one thing, the numbers are unnecessarily hard to remember, especially since the values differ from those used with VimwikiUISelect. This makes creating the links correctly more difficult, and it makes the links more difficult to understand. "Which one is wiki7?" 

But the biggest thing is that they are brittle. If I add a wiki to my wiki list, I break all my interwiki links that point to wikis later in the list. Same if I remove a wiki. Or if I reorder my wikis. Using the schema with a number means that any changes to g:vimwiki_list may break an arbitrary number of interwiki links. 

The solution I propose is to allow users to provide a name for each wiki, if they desire, and then to allow use of that name in an interwiki link syntax. Essentially, this is like DNS. Each wiki has a numeric address, yes, but for human-computer interaction using a meaningful (domain or wiki) name is usually superior. It is more understandable, and it provides a level of indirection allowing the underlying numeric address to change without needing to fix up any links.

At first I just implemented this using VimwikiLinkHandler:

```
" VimwikiHandler for named interwiki links 
function! VimwikiLinkHandler(cmd, link)
  " Use Vim to open named interwiki files via the 'wn.<name>:' scheme.  
  " For example [[wn.spanish:cognates]] 
  let lnk = a:link
  let wnscheme = '^wn\.\([a-zA-Z0-9]\+\):\(.*\)' 
  let wnmatch = matchlist(lnk, wnscheme)
  if len(wnmatch) >= 2 && wnmatch[1] != '' && wnmatch[2] != '' 
    let wname = wnmatch[1]
    let idx = 0
    while idx < len(g:vimwiki_list)
      let wiki = g:vimwiki_list[idx] 
      if has_key(wiki, 'name')
        if wiki.name == wname 
          let lnk = 'wiki' . idx . ':' . wnmatch[2] 
          let lnk = substitute(lnk, VimwikiGet('ext', idx).'$', '', '')
          call vimwiki#base#open_link(a:cmd, lnk)
          return 1
        endif
      endif
      let idx += 1
    endwhile
    echom 'Vimwiki Error: No wiki found with name "' . wiki . '"' 
    return 0
  endif
  return 0
endfunction  
```

[...]

PR #675, opened 2019-04-17
Add wikilocal option to assign a name to each wiki.

### LINKED ISSUE #612: Vimwiki menu only shows one entry if wiki paths all end with same folder name
I have about 12 different vim wikis and they all use the same ending folder name: `vimwiki`. (Make it easier for other tools to find them throughout the file system.)

For example:
```
/project1/vimwiki/
/project2/vimwiki/
/projectN/vimwiki/
```

The problem is that when vimwiki computes the index menu, it appears to only use the final directory name and since they all match, I wind up with a single entry (which activates the last wiki in my list of wikis).

One solution would be to enumerate matching names when mapping them to the index: `vimwiki1`, `vimwiki2`, ... 

A more useful approach would be to use the next higher level in the path when names match (`projectN` in my example). Of course those could theoretically match too in which case you'd have to move yet one more level up.  It seems reliable to assume that the paths _will_ differ at some level or they couldn't refer to different locations on the file system.

ranebrown:
Add `vimwiki-option-name` to assign a per wiki name as outlined in #477. This adds the ability to link between wikis with links of the form `[[wn.name:link]]`. This also updates `:VimwikiUISelect` and `g:vimwiki_menu` output to utilize the wiki name. 

Fixes #612

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -133,6 +133,7 @@ function! vimwiki#base#resolve_link(link_text, ...)
 
   let scheme = matchstr(link_text, '^\zs'.vimwiki#vars#get_global('rxSchemes').'\ze:')
   if scheme == ''
+    " interwiki link scheme is default
     let link_infos.scheme = 'wiki'.source_wiki
   else
     let link_infos.scheme = scheme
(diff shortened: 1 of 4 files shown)


hq6:
@ranebrown , This appears to work very well as long as wiki names are distinct. 
However, if I give two wikis the same name, they get collapsed in the menu.

While I would consider this to be user error, I think it'd be nice if we could warn them about this either in the documentation or by showing a warning at load time if we detect a duplicate name.

PR #686, opened 2019-04-25
New option g:vimwiki_key_mappings to enable/disable key mappings.

### LINKED ISSUE #425: if !hasmapto VimwikiTableAlign{W,Q}
I'd like to use **vim-sentencewrap**'s mapping `gw` which breaks only at punctuation marks .;:?! in a vimwiki file

but vimwiki remaps it regardless of what I have in `my/ftplugin/vimwiki.vim`.

https://github.com/vimwiki/vimwiki/blob/dev/ftplugin/vimwiki.vim#L612
```vim
nnoremap <buffer> gqq :VimwikiTableAlignQ<CR>
nnoremap <buffer> gww :VimwikiTableAlignW<CR>
```

why is this not enclosed in an if statement?

### LINKED ISSUE #671: Make vimwiki textobjects optional / configurable
Hi, 

I have a mapping for [inner line and outer line](https://github.com/kana/vim-textobj-line) _al / il_ using with [vim-sandwich](https://github.com/machakann/vim-sandwich), which conflicts with the list item text-object.

**ftplugin/vimwiki.vim**
```
(...)
" ------------------------------------------------
" Text objects
" ------------------------------------------------

onoremap <silent><buffer> ah :<C-U>call vimwiki#base#TO_header(0, 0, v:count1)<CR>
vnoremap <silent><buffer> ah :<C-U>call vimwiki#base#TO_header(0, 0, v:count1)<CR>

onoremap <silent><buffer> ih :<C-U>call vimwiki#base#TO_header(1, 0, v:count1)<CR>
vnoremap <silent><buffer> ih :<C-U>call vimwiki#base#TO_header(1, 0, v:count1)<CR>

onoremap <silent><buffer> aH :<C-U>call vimwiki#base#TO_header(0, 1, v:count1)<CR>
vnoremap <silent><buffer> aH :<C-U>call vimwiki#base#TO_header(0, 1, v:count1)<CR>

onoremap <silent><buffer> iH :<C-U>call vimwiki#base#TO_header(1, 1, v:count1)<CR>
vnoremap <silent><buffer> iH :<C-U>call vimwiki#base#TO_header(1, 1, v:count1)<CR>

onoremap <silent><buffer> a\ :<C-U>call vimwiki#base#TO_table_cell(0, 0)<CR>
vnoremap <silent><buffer> a\ :<C-U>call vimwiki#base#TO_table_cell(0, 1)<CR>

onoremap <silent><buffer> i\ :<C-U>call vimwiki#base#TO_table_cell(1, 0)<CR>
vnoremap <silent><buffer> i\ :<C-U>call vimwiki#base#TO_table_cell(1, 1)<CR>

onoremap <silent><buffer> ac :<C-U>call vimwiki#base#TO_table_col(0, 0)<CR>
vnoremap <silent><buffer> ac :<C-U>call vimwiki#base#TO_table_col(0, 1)<CR>

onoremap <silent><buffer> ic :<C-U>call vimwiki#base#TO_table_col(1, 0)<CR>
vnoremap <silent><buffer> ic :<C-U>call vimwiki#base#TO_table_col(1, 1)<CR>

onoremap <silent><buffer> al :<C-U>call vimwiki#lst#TO_list_item(0, 0)<CR>
vnoremap <silent><buffer> al :<C-U>call vimwiki#lst#TO_list_item(0, 1)<CR>

onoremap <silent><buffer> il :<C-U>call vimwiki#lst#TO_list_item(1, 0)<CR>
vnoremap <silent><buffer> il :<C-U>call vimwiki#lst#TO_list_item(1, 1)<CR>
(...)
```
An option to change or disable the textobjects would be nice.

ranebrown:
All key mappings have the option of being disabled but the <Plug>
definitions are still available for user customization. Key sequences
are no longer overwritten if a mapping already exists. Fixes #671,
fixes #425.

### DIFF
--- a/autoload/vimwiki/vars.vim
+++ b/autoload/vimwiki/vars.vim
@@ -166,6 +166,12 @@ function! s:read_global_settings_from_user()
         \ 'hl_headers': {'type': type(0), 'default': 0, 'min': 0, 'max': 1},
         \ 'html_header_numbering': {'type': type(0), 'default': 0, 'min': 0, 'max': 6},
         \ 'html_header_numbering_sym': {'type': type(''), 'default': ''},
+        \ 'key_mappings': {'type': type({}), 'default':
+        \   {
+        \     'all_maps': 1, 'global': 1, 'headers': 1, 'text_objs': 1,
+        \     'table_format': 1, 'table_mappings': 1, 'lists': 1, 'links': 1,
+        \     'html': 1, 'mouse': 0,
+        \   }},
         \ 'list_ignore_newline': {'type': type(0), 'default': 1, 'min': 0, 'max': 1},
         \ 'text_ignore_newline': {'type': type(0), 'default': 1, 'min': 0, 'max': 1},
         \ 'links_header': {'type': type(''), 'default': 'Generated Links', 'min_length': 1},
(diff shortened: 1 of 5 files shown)


ranebrown:
@hq6 This could use a second set of eyes. I tested out the key mappings but want to make sure nothing was missed. Some of the default key mappings don't work for me like `<S-CR>` so I had to modify those for testing and then change them back.

hq6:
@ranebrown , Taking a look, but please fix conflicts when you get a chance.

ranebrown:
I'll wait until you provide any other feedback since the conflict is just the changelog and any other PRs that get merged first will create a new conflict.

autoload/vimwiki/vars.vim
```diff
@@ -248,6 +254,63 @@ function! s:normalize_global_settings()
...
+    let g:vimwiki_global_vars.key_mappings.table_mappings = 0
+    let g:vimwiki_global_vars.table_mappings = 0 " kept for backwards compatibility
+    let g:vimwiki_global_vars.key_mappings.lists = 0
+    let g:vimwiki_global_vars.key_mappings.links = 0
+    let g:vimwiki_global_vars.key_mappings.html = 0
+    let g:vimwiki_global_vars.key_mappings.mouse = 0
+    let g:vimwiki_global_vars.use_mouse = 0 " kept for backwards compatibility
+  endif
+
+  " backwards compatibility checks
```
hq6: Do we want to add a note into the code itself to indicate that these old options are deprecated and remind ourselves to remove them after a release or two?

hq6: Also, since the documentation describes the new options instead of the old ones, it may be better to only overwrite new with old if both of the following conditions are true.
 * The old option isn't its default value
 * The new option is its default value.

If the new option is also set, I would favor respecting the new option for consistency with the documentation.

doc/vimwiki.txt
```diff
@@ -3326,6 +3408,8 @@ https://github.com/vimwiki-backup/vimwiki/issues.
 2.5 (in progress)~
 
 New:~
+    * PR #686: New option |g:vimwiki_key_mappings| that allow key mappings to
```
hq6: Do we also want to mention "Key sequences are no longer overwritten if a mapping already exists"?

autoload/vimwiki/vars.vim
```diff
@@ -166,6 +166,12 @@ function! s:read_global_settings_from_user()
         \ 'hl_headers': {'type': type(0), 'default': 0, 'min': 0, 'max': 1},
         \ 'html_header_numbering': {'type': type(0), 'default': 0, 'min': 0, 'max': 6},
         \ 'html_header_numbering_sym': {'type': type(''), 'default': ''},
+        \ 'key_mappings': {'type': type({}), 'default':
+        \   {
+        \     'all_maps': 1, 'global': 1, 'headers': 1, 'text_objs': 1,
```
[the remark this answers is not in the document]
ranebrown: The main reason for this was to maintain compatibility with the old table_mappings option but I think it still makes sense to keep the two options since it is more common for the `<Tab>` insert mode mappings to cause a conflict.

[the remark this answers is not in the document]
ranebrown: Other than that I tried to base the groups on similar functionality.

doc/vimwiki.txt
```diff
@@ -548,8 +554,11 @@ gw1                     tested. If the alignment of the current line differs,
                         To remap: >
                         :nmap <Leader>j <Plug>VimwikiDiaryNextDay
 <
+Mouse mappings          *vimwiki_mouse*
+
+These mappings are disabeld by default.
```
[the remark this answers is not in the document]
ranebrown: Fixed

autoload/vimwiki/vars.vim
ranebrown: I agree, done.

doc/vimwiki.txt
ranebrown: Done

ranebrown:
@hq6 I made changes as you suggested and fixed the conflicts. Should be good to go. Thanks!

autoload/vimwiki/vars.vim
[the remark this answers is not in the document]
hq6: I actually meant documenting it in the code rather than in this PR, so that future readers of this code will be able to tell why things were divided this way. However, I think it's merge-able as is because if someone was curious they could `git blame` and find this PR.

Issue #683, opened 2019-04-24
shift-enter and control-enter not working?


sdondley:
Branch: dev
Revision: 2b6fa27
Date: 2019-04-22 07:39:11 -0700

The `shift-enter` and `control-enter` commands don't open the new file up in a split which I think is the expected behavior if I'm interpreting the documentation correctly. Instead, the link is just opened in a brand new buffer by itself.

My .vimrc:

```
syntax on
filetype plugin on
set nocompatible
set background=dark
set cindent
set tabstop=2
set shiftwidth=2
set expandtab
nnoremap sv :source $MYVIMRC<CR>

execute pathogen#infect()
execute pathogen#helptags()
```

I have removed all other plugins.

sdondley:
I just read the help docs. It said it may not work in all terminals and to try remapping. I'm using iTerm2.

hq6:
@sdondley, I think the problem is that many terminals capture these keybindings and try to interpret them, rather than passing them onto Vim. 
I use `gnome-terminal` and it has the same problem, but remapping seems to work fine. 
For example, I tried the following mapping in my `.vimrc` and it did the correct behavior.

```
nmap  <Leader>ws <Plug>VimwikiSplitLink
```

Please us know if that helps, but this is ultimately not a Vimwiki issue but rather a terminal issue. We could change the default mappings, but I do not know of any way to ensure that they work on all terminals.

sdondley:
Yeah, I just did a remap. If you'd like, I can submit a patch to document this in the readme. May save people some frustration.

sdondley:
I made a change to the changelog with a PR # of 683, this issue. This is actually PR #687, though. Not sure if I did things correctly. Please advise.

PR #687, opened 2019-04-25
improve documentation of key bindings

sdondley:
For #683

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,13 @@ For other syntax elements, see `:h vimwiki-syntax`
 
 ## Key bindings
 
-Normal mode:
+### Normal mode
+
+**Note:** your terminal may prevent capturing some of the default bindings
+listed below. See `:h vimwiki-local-mappings` for suggestions for alternative
+bindings if you encounter a problem.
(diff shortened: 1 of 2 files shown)


ranebrown:
@sdondley It doesn't look like you made any changes to the README. You can make your suggested changes and commit them for us to review.

The documentation accessed with `:h vimwiki` does have a note about the key mapping not working in all terminals ([here](https://github.com/vimwiki/vimwiki/blob/2b6fa274a8db8463d494dffea6fe9e98796f065d/doc/vimwiki.txt#L289) and [here](https://github.com/vimwiki/vimwiki/blob/2b6fa274a8db8463d494dffea6fe9e98796f065d/doc/vimwiki.txt#L281)).

Your name in the list of contributors should also be placed at the bottom of the list.

sdondley:
OK, make a new pull request. Yes, I realize it's in the documentation but it will save people time and aggravation if they have a head's up about it and know where in the documentation to go to resolve the issue.

ranebrown:
Looks good, thanks!

Issue #685, opened 2019-04-24
`s:check_users_value()` returns help suggestions that do not exist for options


agude:
For example, when I mis-set `diary_rel_path` (as in #684), it gives the error:
```
Vimwiki Error: The provided value '' of the option 'diary_rel_path' in g:vimwiki_list must have at least 1 character(s) but has 0. See ':h g:vimwiki_diary_rel_path'.
```

But running `:h g:vimwiki_diary_rel_path` gives the error:
```
E149: Sorry, no help for g:vimwiki_diary_rel_path
```

Instead, it should suggest:
```
:h vimwiki-option-diary_rel_path
```

ranebrown:
Thanks for pointing this out. There is a common function used for printing the error messages that needs updated for the wikilocal options. If you can make the change as part of your PR for #684 that would be great.

agude:
I'll give this one a go, but I suspect it will take me a bit to wrap my head around it.

PR #689, opened 2019-04-26
Change diary_rel_path minimum length to 0

### LINKED ISSUE #684: `diary_rel_path` has minimum length 1, but could be 0
Recently some error checking was added to the parsing of the user settings: https://github.com/vimwiki/vimwiki/commit/fdc367f7250989e78ccbd6454fade109e008cdaa

Previously I had been using `'diary_rel_path': '',` to put my diary files in the same directory as the rest of my wiki files, but now this breaks because `diary_rel_path` has a minimum length of 1. See here: https://github.com/vimwiki/vimwiki/blob/be793e28bea4ac30c19c15c4db3fd42d26a5debe/autoload/vimwiki/vars.vim#L253

This leads to the following error: 
```
Vimwiki Error: The provided value '' of the option 'diary_rel_path' in g:vimwiki_list must have at least 1 character(s) but has 0. See ':h g:vimwiki_diary_rel_path'.
```

A work around is I can set `'diary_rel_path': '.',`, but it seems like there is no reason to force this variable to have a set number of characters.

agude:
This fixes #684.

### DIFF
--- a/autoload/vimwiki/vars.vim
+++ b/autoload/vimwiki/vars.vim
@@ -265,7 +265,7 @@ function! s:populate_wikilocal_options()
         \ 'custom_wiki2html_args': {'type': type(''), 'default': ''},
         \ 'diary_header': {'type': type(''), 'default': 'Diary', 'min_length': 1},
         \ 'diary_index': {'type': type(''), 'default': 'diary', 'min_length': 1},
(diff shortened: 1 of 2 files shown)


agude:
I think I'll leave off fixing #685 for now. It's going to take a little more reading to figure out the right fix.

PR #692, opened 2019-04-29
Allow user to use a custom function for foldexpr option

sdondley:
Consider this a conversation starter. I'll post another commit for documentation if this is a go.

The patch is pretty straightforward. It adds a new user boolean variable, `custom_fold_func`. If true, it will call `VimwikiFoldLevelCustom` which the user will need to supply.

I created this because I did not like how vimwiki folded blank lines into the folds. I created this custom function to prevent the last blank lines from getting folded (note this function is not part of the patch, it's just here for reference for others who may want to use it). It works for markdown syntax:

```
let g:lastnum = 0
function! VimwikiFoldLevelCustom(lnum)
    if getline(a:lnum) =~? '\v^\s*$'
	let g:lastnum = -1
        return '-1'
    endif
    let last = g:lastnum
    let num_of_pounds = strlen(matchstr(getline(a:lnum), '^#\+'))
    if (num_of_pounds > 0)
            let g:lastnum = num_of_pounds
	    if last == -1
		    return '>' . num_of_pounds
	    endif
	    return num_of_pounds
    endif
    return g:lastnum
endfunction
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,13 @@ For other syntax elements, see `:h vimwiki-syntax`
 
 ## Key bindings
 
-Normal mode:
+### Normal mode
+
+**Note:** your terminal may prevent capturing some of the default bindings
+listed below. See `:h vimwiki-local-mappings` for suggestions for alternative
+bindings if you encounter a problem.
+
+#### Basic key bindings
 
  * `<Leader>ww` -- Open default wiki index file.
  * `<Leader>wt` -- Open default wiki index file in a new tab.
@@ -192,7 +198,10 @@ Normal mode:
  * `<Tab>` -- Find next wiki link
  * `<Shift-Tab>` -- Find previous wiki link
 
-For more keys, see `:h vimwiki-mappings`
+#### Advanced key bindings
+
+Refer to the complete documentation at `:h vimwiki-mappings` to see many
+more bindings.
 
 ## Commands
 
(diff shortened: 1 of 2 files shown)


ranebrown:
I think this is a reasonable request. Instead of creating a new variable it would make sense to use a [callback](https://github.com/vimwiki/vimwiki/blob/61d93fde7ffc4298e606871b09d72077863afe87/plugin/vimwiki.vim#L231-L249) like is done for some of the other functions. Then an extra option could be added to `g:vimwiki_folding` instead of adding a new option.

The example function should probably be placed in the documentation similar to what is done for the other callbacks.

Make sure to rebaseline against the latest `dev`.

sdondley:
OK, sounds good.

Yeah, sorry about the old commits in there. What's the command for "rebaselining?" I don't use git much collaboratively so I get a little confused when working on new projects. Each seems to have its own set of protocols for using git. I wrongly assumed that since my last pull request was approved, it would not get submitted.

ranebrown:
No worries. On your fork you need to [setup a remote](https://help.github.com/en/articles/configuring-a-remote-for-a-fork) that points to Vimwiki/dev. Then you can just `git pull upstream/dev`.

sdondley:
OK, after looking at this closer, I don't think a patch is necessary. A user can set `g:vimwiki_folding` to 'custom' and set the necessary fold options in their .vimrc file.  I think all that is needed a little more documentation, perhaps with my example, in the `g:vimwiki_folding` section of the docs.

Let me know what you think.

sdondley:
Yeah, the `6. Folding/Outline` portion of the main documentation hasn't been updated in at least 6 years. I'll update it.

hq6:
I have responded on your other PR and this one looks like it is no longer necessary, so I am closing it.

Issue #445, opened 2018-02-26
When moving between files, vim/vimwiki displays info about the changes


guakamole:
Hello (again),

I hope that I am being helpful (and not a burden) by reporting these issues, I know that you have been working hard on refactoring the code and have recently integrated some of this refactoring. Thank you again for all your work and thank you to the contributors :+1: 

## Bug or feature?

When navigating between wiki pages (using `Enter` on an internal link and `Backspace` on a wiki page), vim/vimwiki displays some stats about the current page and/or updates  if the page has been modified.

## Steps to reproduce

```
vim -c VimwikiIndex
# navigate to next internal link (foobar), hit Enter
# go to the end of page, insert a word
# return to normal mode, hit Backspace
```

```
"~/vimwiki/foobar.md" 150L, 6789C written
"~/vimwiki/index.md" 372L, 134501C
Press ENTER or type command to continue
```

EinfachToll:
Should be fixed now.

PR #698, opened 2019-05-04
Remove awa check triggering silent file edits.

jpmor:
This is a change to `vimwiki#base#edit_file`.  Currently `if &autowriteall && !&hidden` are set, commands are executed silently.  The justification in comments for this is:
```in this case, the file is saved before switching to the new buffer. This causes Vim to show two messages in the command line which triggers the annoying hit-enter prompt. Solution: show no messages at all.``` 

Here's the original issue that lead to this being added: https://github.com/vimwiki/vimwiki/issues/445

I've opened a PR to remove this because (1.) The bug described in that issue and in the above comment is not occurring anymore and (2.) it's useful to see filename and cursor position displayed when switching links, and that is lost with `silent execute`.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -826,22 +826,16 @@ function! vimwiki#base#edit_file(command, filename, anchor, ...)
   " This hack is necessary because apparently Vim messes up the result of
   " getpos() directly after this command. Strange.
   if !(a:command ==# ':e ' && vimwiki#path#is_equal(a:filename, expand('%:p')))
-    if &autowriteall && !&hidden  " in this case, the file is saved before switching to the
-      " new buffer. This causes Vim to show two messages in the command line which triggers
-      " the annoying hit-enter prompt. Solution: show no messages at all.
-      silent execute a:command fname
-    else
-      try
-        execute a:command fname
-      catch /E37:/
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ranebrown** [comment]
@jpmor my concern with this change is that the problem might still occur on older Vim versions. I think it is also dependent on your `cmdheight` setting and possibly other settings. 

Have you tested different Vim versions and using a minimal `.vimrc`?

**Comment by jpmor** [comment]
Hey @ranebrown - I tried the following `.vimrc` where `~/vimwiki` has the conditional removed:
```
call plug#begin('~/.vim/plug')
  Plug '~/vimwiki'
call plug#end()```
I gave this a try with 7.4, 8.0, 8.1  and was not able to reproduce the "hit-enter" bug.

ranebrown:
I also tested and didn't get the hit enter prompt. @jpmor can you update the changelog?

jpmor:
@ranebrown just updated changelog.

ranebrown:
@guakamole can you confirm removing this check doesn't cause problems since you opened the original issue? I want to make sure it isn't specific to certain settings since we have not been able to reproduce the behavior described in #445.

ranebrown:
@jpmor can you update the changelog and move this to the removed section? I was hoping to get some more feedback but doesn't look like that is going to happen.

jpmor:
Hi @ranebrown - updated the changelog again and move to the removed section.

Issue #696, opened 2019-05-03
Export Visual Selection


jupart:
I think this would be an enhancement. Adding the ability to export a visual selection to HTML would be very nice. I plan tasks for myself and others in vimwiki. Having a way to select, export, and share a project (grouped by a heading) or a group of tasks would be a time saver.

hq6:
I agree that seems like a useful feature. 
Would you be interesting in creating a PR for it?

jupart:
Yes, I'll look into it, but I wanted to create an issue first.

PR #701, opened 2019-05-06
Add VimwikiVisualSelection2HTML, need to factor out copied code

jupart:
For #696 .

TODO:
- Factor out duplicate code
- Add VimwikiVisualSelection2HTML to doc/vimwiki.txt
- Incorporate additional feedback


ranebrown:
@hq6 are you going to have time to test and review this one?

ranebrown:
@jupart I took a quick look at this and it seems to work with default settings but fails when using `custom_wiki2html`. Is supporting that option possible? It also might be a good idea to modify the output filename to include something like `_visual`. 

Are you still working this? It looks like you have a few open todos.

ranebrown:
Closing since there are open todos and it isn't being worked.

Issue #702, opened 2019-05-07
Make remapping documentation more accessible to newer vim users


sdondley:
As a rather amateur Vim user, I found Section 3.1 on "Global mappings" to be somewhat confusing and has other problems:

1. I had no idea what `<Plug>` was. Though I had seen this before, I didn't know what to do with it or why it was important.
2. To remap a key, the documentation suggests that `:nmap` be used. However, it's best practice to use `nnoremap`
3. The documentation seems to provide a suggestion for another key for remapping. Example:

`:nmap <Leader>w <Plug>VimwikiIndex`

However, a newer vim user may not realize they can make the {lhs} with anything they want. It doesn't have to include the `<Leader>` key or the suggested remap (`w` in the above example).

Suggestions for improvement:

1. Place a clear tutorial on how to perform remaps at the top of the section
2. Remove confusing documentation

Proposed patch forthcoming.

ranebrown:
> To remap a key, the documentation suggests that :nmap be used. However, it's best practice to use nnoremap

The `<Plug>` mappings need to be recursive to work properly. The actual definitions use the `nnoremap` versions if you look in `ftplugin/vimwiki.vim`. Anything that doesn't use `<Plug>` should use the `nnoremap` version as you suggested.

sdondley:
Yeah, figured that out afterwards and made a new commit to address my error.

PR #703, opened 2019-05-07
improve global mapping documentation on remapping

sdondley:
for #702

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -110,12 +110,13 @@ There are global and local mappings in Vimwiki.
 ------------------------------------------------------------------------------
 3.1. Global mappings                                 *vimwiki-global-mappings*
 
-Below is a list of all default mappings provided by Vimwiki. Vimwiki respects
-pre-existing key mappings created by you or other plugins and will not
-overwrite them.
+Below is a list of all default global key mappings provided by Vimwiki. As
+global settings, they work in all vim sessions no matter what filetype you
+have open. Vimwiki respects pre-existing global mappings created by you or
+other plugins and will not overwrite them.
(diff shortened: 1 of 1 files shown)


sdondley:
I need to fix this patch. Don't commit it yet.

doc/vimwiki.txt
```diff
@@ -229,8 +256,8 @@ See also |:VimwikiMakeYesterdayDiaryNote|
         3<Leader>w<Leader>m opens diary wiki-file for tomorrow in the third
         wiki from |g:vimwiki_list|.
         etc.
-        To remap: >
-        :nmap <Leader>dm <Plug>VimwikiMakeTomorrowDiaryNote
+
+        Remap command: `<Plug>VimwikiMakeTomorrowDiaryNote`
 <
```
ranebrown: This character can be removed

doc/vimwiki.txt
```diff
@@ -110,47 +110,71 @@ There are global and local mappings in Vimwiki.
...
+where
+        `{map}` is the new key sequence of your choosing
+        `{command}` is the Vimwiki command you are remapping
+
+So, for example, to remap the |:VimwikiIndex| mapping, you'd do something
+like: >
+        :nmap <Leader>wx <Plug>VimwikiIndex
+
+Note that the "map" command is needed as "noremap" will not work. This is fine
+in this case because vim's `<Plug>` key code will make an inadvertent remapping
```
ranebrown: Maybe slightly modify this to say that noremap can't be used because for the mapping to be properly setup Vim needs to recursively expand the right hand side of the mapping to get the `<Plug>` definition.

PR #704, opened 2019-05-08
Doc update for Section 3.2, Local mappings

sdondley:
This is an update to documentation to make it consistent with #703


ranebrown:
Waiting on #699 since there might need to be some updates depending on the decision there.

ranebrown:
@sdondley can you update this since the behavior was changed to overwrite other mappings for the buffer specific mappings? Maybe add a note that points users to `g:vimwiki_key_mappings`.

Issue #680, opened 2019-04-21
List folding hides next list items


jmolero:
Hi. 

There's a problem when using the list folding. It folds list more than it should and that can hide a lot of information in long lists.

I followed the example in the docs and it works well in that case. But if you take the following example that doesn't have `  * [ ] Do substuff 1.3`:

```
* [ ] Do stuff 1
    * [ ] Do substuff 1.1
    * [ ] Do substuff 1.2
        * [ ] Do substuff 1.2.1
        * [ ] Do substuff 1.2.2
* [ ] Do stuff 2
* [ ] Do stuff 3
``` 

Hitting zM makes it fold like this: 

```
= My current task =
* [ ] Do stuff 1 [6] ------------------
* [ ] Do stuff 3
```

I think it should look like this instead:

```
= My current task =
* [ ] Do stuff 1 [5] ------------------
* [ ] Do stuff 2
* [ ] Do stuff 3
```

This seems to  happen because the first sublist doesn't end in the same indent level.

I tried with the the dev branch but it has the same behavior.

I'm using this version: 

```
Branch: master
Revision: be793e2
Date: 2019-04-20 14:13:17 +0200
```

hq6:
Thanks for reporting this!
I can confirm that this behavior reproduces and definitely seems broken.
We will look into this eventually, but if you are interested in contributing and see it fixed faster, a PR is quite welcome!

ranebrown:
Has anyone checked if this behavior has always been present or was introduced in one of the more recent releases?

hq6:
@ranebrown Not yet, but I can bisect it and report back.

hq6:
Wow, this issue was introduced in c6e110a368562c30ac3126cec6cccd46efee8eb6 from 2013. 
The fact that we haven't found it until now seems to imply that not many people use `g:vimwiki_folding = 'list'`.

PR #705, opened 2019-05-08
fix list folding bug #680

sdondley:
For #680 

I think this should fix it. Seemed to work in my limited manual testing.

### DIFF
--- a/autoload/vimwiki/lst.vim
+++ b/autoload/vimwiki/lst.vim
@@ -1689,7 +1689,7 @@ function! vimwiki#lst#fold_level(lnum)
     if child_item.type != 0
       return 'a1'
     elseif next_item.type == 0
-      return 's1'
+      return '<' . foldlevel(parent_item.lnum)
     endif
   endif
   return '='
--- a/autoload/vimwiki/lst.vim
+++ b/autoload/vimwiki/lst.vim
@@ -1689,13 +1689,9 @@ function! vimwiki#lst#fold_level(lnum)
     if child_item.type != 0
       return 'a1'
     elseif next_item.type == 0
-<<<<<<< Updated upstream
-      return '<' . foldlevel(parent_item.lnum)
-=======
         let c_indent = indent(a:lnum) / &shiftwidth
         let n_indent = indent(a:lnum+1) / &shiftwidth
         return 's' . (c_indent - n_indent)
->>>>>>> Stashed changes
     endif
   endif
   return '='
--- a/autoload/vimwiki/lst.vim
+++ b/autoload/vimwiki/lst.vim
@@ -1689,7 +1689,13 @@ function! vimwiki#lst#fold_level(lnum)
     if child_item.type != 0
       return 'a1'
     elseif next_item.type == 0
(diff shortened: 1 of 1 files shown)


ranebrown:
Your change doesn't seem to work correctly. The example from the linked issue doesn't fold at all for me. I think this needs some more thorough testing.

sdondley:
Can you confirm you have `g:vimwiki_folding` set to "list"?

sdondley:
Here's a gif movie of it working for me.

fold_recording

ranebrown:
Seem to be related to having multiple lists or maybe mixed checkboxes. I didn't really dig into it much. With the below two lists nothing in the second one is folded.
```
- [ ] list 
    - [ ] blah
    - [ ] another item
        - info
        - more stuff
        - link

* [ ] Do stuff 1
    * [ ] Do substuff 1.1
    * [ ] Do substuff 1.2
        * [ ] Do substuff 1.2.1
        * [ ] Do substuff 1.2.2
* [ ] Do stuff 2
* [ ] Do stuff 3
```

sdondley:
Ok, see last patch. I tried to avoid using differences in indentation because, according to line 485 of the "lst.vim" file: "child items do not necessarily have the same indent, i.e. level".  But I don't see an obvious way to avoid using the indent level. It's the only way to determine how many fold levels less the current line is from the previous line. I discovered It's impossible to get the parent level for an item with `foldlevel(parent.lnum)` because, according to the documentation, vim returns -1 until all the fold levels have been determined.

Also, I should note, the vim docs recommend against using 's1', 'a1', 's2' to add and subtract fold levels as it's slow.

But this is at least an improvement and it's a lot better than ripping the guts out of how folds are calculated.

ranebrown:
Thanks for the quick turnaround on this! It does seem to be working better now. I noticed that the two lists get folded together even if there is text in between them. That seems to be the same behavior prior to your change but I don't know that it makes sense. I would think separate lists should be separate folds. What do you think?

sdondley:
I'm seeing two different lists fold. See gif.

ranebrown:
Using this text and closing the fold on 'bar' folds everything below that:
```
= Page =

- bar
   - bar 1
   - bar 2
      * baz

== section ==

- [ ] Do stuff 1
    - [ ] Do substuff 1.1
    - [ ] Do substuff 1.2
        - [ ] Do substuff 1.2.1
        - [ ] Do substuff 1.2.2
- [ ] Do stuff 2
- [ ] Do stuff 3
```

It does work correctly in certain situations but I'm not sure why.

sdondley:
You have to be sure your tabstop and and shiftwidth are the same and that each line is indented by that amount. If they are, it works:

sdondley:
In you sample, the sub-bullets in the first list are only indented by 3. It's probably possible to tweak the indentation calculation to try to "round" which tab stop the line is at. But 90% of people are just going to use regular tab stops. If they don't, they can write their own folding function.

ranebrown:
Ah, makes sense. I just had some random stuff pasted in the file and didn't notice the different indentation.

PR #710, opened 2019-05-16
VimWiki2HTML -> lowercase 'w' in readme

AresMegaGlobal:
# Pull Request Checks

It should be Vimwiki2HTML, right?


ranebrown:
Yep, thanks!

PR #711, opened 2019-05-20
Allow forcing VimwikiAll2HTML with !

gpanders:
This simple change allows the user to supply a `!` to the `:VimwikiAll2HTML` command which forces vimwiki to recompile wikis to HTML even if they are not marked as out of date.

This is useful in the case where someone updates either the CSS file or the HTML template file (in which case the actual wiki files will not be marked out-of-date) and wants to recompile all of their wikis.


ranebrown:
@gpanders Looks good to me. Can you update the changelog and then I'll merge?

gpanders:
@ranebrown Done.

PR #718, opened 2019-05-31
Fix not normalized path separators on Windows

blyoa:
Currently, `vimwiki#path#path_norm` calls `resolve` after all path normalization is finished, but when a path is resolved by `resolve` on Windows, "/" in the path changes back to "\\". This leads `vimwiki#base#find_wiki` to fail to find a registered wiki in `g:vimwiki_list`.

### DIFF
--- a/autoload/vimwiki/path.vim
+++ b/autoload/vimwiki/path.vim
@@ -38,11 +38,12 @@ endfunction
 function! vimwiki#path#path_norm(path)
   " /-slashes
   if a:path !~# '^scp:'
-    let path = substitute(a:path, '\', '/', 'g')
+    " ensure that we are not fooled by a symbolic link
+    let path = resolve(a:path)
+    let path = substitute(path, '\', '/', 'g')
     " treat multiple consecutive slashes as one path separator
     let path = substitute(path, '/\+', '/', 'g')
-    " ensure that we are not fooled by a symbolic link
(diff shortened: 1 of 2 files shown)


ranebrown:
@blyoa was there a specific command that you were seeing issues with? I have done some testing for other issues on Windows and didn't have problems opening a wiki.

blyoa:
@ranebrown, I'm sorry for confusing you. This issue no longer occurs. I will close this issue after this comment.

I found this issue when I was using Vim 8.1.1436. In Vim 8.1.1436, `resolve` replaced the slashes of any directory path to backslash, even though the directory path was not a symbolic link. Although the path used in `vimwiki#base#find_wiki(path)` was already resolved in `s:normalize_wikilocal_settings`, when `vimwiki#base#find_wiki(path)` called `vimwiki#path#path_norm(path)`, which called `resolve`, the slashes of `path` were replaced to backslashes again. As a result, `vimwiki#base#find_wiki(path)` could not find wiki files at least on Vim 8.1.1436–8.1.1471.

I tried to reproduce this issue on the latest Vim 8.1.1512, but the issue no longer occurred. I think this issue was solved by the patch 8.1.1473: "new resolve() implementation causes problem for plugins Update runtime files."

I'm sorry for the wrong issue report. I will close this issue.

PR #735, opened 2019-07-30
Fix off-by-one error in get_next_line and get_prev_line

### LINKED ISSUE #407: Fenced code block in list breaks list in Markdown
````
- Item 1
- Item 2
   ```
   echo "hello world"
   ```
- Item 3
````

In the above, `Item 3` is no longer connected to the list containing `Item 1` and `Item 2`. So renumbering the list/changing list style no longer works.

### PR DESCRIPTION
**Pull request by simias**
I believe this fixes #407.

The problem as far as I can tell is that in the following example:

```
1 * [ ] Test
2     + [ ] Foo
3     ```c
4     printf("Hello, World\n")
5     ```
6     + [ ] Bar
7     + [ ] Baz
```

If `get_next_line` was used to get the next item after Foo it would correctly find
the `rxPreStart` symbol (```` ``` ````) and skip over it. Then it would loop until it
encountered the `rxPreEnd` symbol (the same ```` ``` ````) and there it would stop.

So far everything is working as expected, except that it would return the
*current* line (the one with the closing ```` ``` ````) instead of the next one. Therefore
the next call to get_next_line would start on the *same* line. Since for
markdown rxPreStart and rxPreEnd are the same sequence it now thinks this this
the start of a new block and keeps reading until it finds ```` ``` ```` again (or the end
of file).

The bug wasn't apparent with vimwiki markup because it uses it uses a different
sequence to begin and end code blocks, so the subsequent call to get_next_line
simply ignored the end-of-block sequence instead of erroneously believing it was the beginning of a new block.

### DIFF
--- a/autoload/vimwiki/lst.vim
+++ b/autoload/vimwiki/lst.vim
@@ -419,10 +419,12 @@ function! s:get_next_line(lnum, ...)
     while cur_ln <= line('$') && getline(cur_ln) !~# vimwiki#vars#get_syntaxlocal('rxPreEnd')
       let cur_ln += 1
     endwhile
-    let next_line = cur_ln
+    let next_line = cur_ln + 1
   else
-    let next_line = nextnonblank(a:lnum+1)
+    let next_line = a:lnum + 1
   endif
+
+  let next_line = nextnonblank(next_line)
 
   if a:0 > 0 && getline(next_line) =~# vimwiki#vars#get_syntaxlocal('rxHeader')
     let next_line = s:get_next_line(next_line, 1)
@@ -440,18 +442,18 @@ endfunction
 "Returns: lnum-1 in most cases, but skips blank lines and preformatted text
 "0 in case of nonvalid line and a header, because a header ends every list
 function! s:get_prev_line(lnum)
-  let prev_line = prevnonblank(a:lnum-1)
+  let cur_ln = a:lnum - 1
 
-  if getline(prev_line) =~# vimwiki#vars#get_syntaxlocal('rxPreEnd')
-    let cur_ln = a:lnum - 1
+  if getline(cur_ln) =~# vimwiki#vars#get_syntaxlocal('rxPreEnd')
     while 1
       if cur_ln == 0 || getline(cur_ln) =~# vimwiki#vars#get_syntaxlocal('rxPreStart')
         break
       endif
       let cur_ln -= 1
     endwhile
-    let prev_line = cur_ln
   endif
+
+  let prev_line = prevnonblank(cur_ln)
 
   if prev_line < 0 || prev_line > line('$') ||
         \ getline(prev_line) =~# vimwiki#vars#get_syntaxlocal('rxHeader')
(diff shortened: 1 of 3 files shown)


simias:
Actually I've just noticed a potential other problem: if you have a blank line after the closing ` ``` ` it also breaks the list. I think the problem is that `nextnonblank` is not called if we're in a block (it's only in the "else" part).
I'll try to provide an updated patch ASAP.

simias:
Okay, it should be good now.

ranebrown:
Thanks! Looks good and seems to be working.

Do you mind taking a stab at creating a Vader test in the new test directory? I'm trying to implement tests to cover as much of the functionality as possible and any know issues. The only thing that should be required is Docker. You will need to build the image and then use the script to run the tests (skip the Vint tests for now). There is a Readme in the test directory and I can help out with any questions. This is pretty new so it isn't a hard requirement for getting this merged but it would definitely help out!

Last thing - please update the changelog indicating this fix.

simias:
Sure, I'll give it a try.

simias:
I added a handful of tests for both vimwiki and markdown mode. I haven't added mediawiki because I'm not super familiar with that syntax and I wasn't sure how to nest code blocks within lists properly. If you can provide a sample I could try to integrate it as well.

I also updated the changelog but I wasn't very inspired, let me know what you think.

simias:
I also reworked the fix itself by the way. The patch is a little more invasive than before but I believe that the resulting code is a tiny bit cleaner and easier to follow. I didn't like duplicating the call to `nextnonblank`.

ranebrown:
Nice! Thanks for giving the tests a stab. I think it would be good to also add in some tests for the `gl` and `gL` variant commands (see `:h vimwiki_glr`. They are used to renumber lists or change the symbol (-,+,*,etc.) for a single list items or the entire list.

For the changelog can you move the update to the 'Fixed:' section and include Issue #407 in the description as well?

simias:
`gll` and friends are very much broken when you have code blocks but I'm not really sure what's the right way to deal with them. Should they indent code as well? Is it even standard to have code that's at a lower level of indentation than the surrounding list?

simias:
By "`gll` and friends" I mean functions dealing with list level changes.

simias:
Nevermind that, renumbering is broken as well. I need to figure that out.

ranebrown:
Looks like modifying the indentation level doesn't work but renumbering and changing the list marker works if the code block is indented at least 1 space past the list level correct?

ranebrown:
> Should they indent code as well? 

I would say yes

ranebrown:
@simias are you going to be able to work on this PR any more? If not I think it is still at a reasonable point to merge and the `gll` commands can be moved to a new issue.

hq6:
@ranebrown Should we go ahead and merge this, or close it?

ranebrown:
Yeah, I just haven't had time to get back to it. I wanted to make sure the test cases cover enough and possibly add some more.

simias:
Sorry for being away for so long, I've been away from my computer for a while and now I'm quite busy. I still plan on finishing this eventually but it's stuck in my backlog for the time being.

I think merging the current state and opening a new issue for the other commands is reasonable, it's not perfect as-is but I don't think it causes any regression (and it does fix a few problems).

ranebrown:
> I still plan on finishing this eventually but it's stuck in my backlog for the time being.

I feel you. This will get merged once I have some more time to look at the test cases.

PR #742, opened 2019-08-21
Register autocmds once

abhinav:
This commit changes the autocmd registration in vimwiki so that
registration happens once rather than once per extension.

Instead of the following,

    autocmd BufEnter *.md call s:setup_buffer_enter()
    autocmd BufEnter *.mdown call s:setup_buffer_enter()

We'll now only run,

    autocmd BufEnter *.md,*.mdown call s:setup_buffer_enter()

This probably has no effect on performance but it makes for a simpler
implementation.

---

This is a pretty minor stylistic PR; feel free to close if this isn't a
change you'd like.

(The changelog wasn't updated because this change has no user-facing
impact.)

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -3457,6 +3457,7 @@ Contributors and their Github usernames in roughly chronological order:
     - Alexander Gude (@agude)
     - Jonny Bylsma (@jbylsma)
     - Shaedil (@Shaedil)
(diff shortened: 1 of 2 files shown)


ranebrown:
Makes sense to remove the for loop if possible. I'll try to look over your changes in a bit more detail this weekend.

Issue #746, opened 2019-08-29
[Question] Markdown: point a file with absolute path


rjshrjndrn:
Hi,
Thank you for this awesome plugin.
I'm trying to figure out one thing but in vain.
`  * [Ansible refactor](/home/rajeshr/Documents/projects/public-devops/README.md)`
How to open the file, in markdown mode ?

ranebrown:
Seems to be a bug, I'll try and look into it but I'm a bit short on time right now.

ranebrown:
As a workaround you could put the below function in your `vimrc` and then open the absolute links with `[link](vfile:/absolute/path/here)`

```vim
function! VimwikiLinkHandler(link)
    let link = a:link
    if link =~# '^vfile:'
        let link = link[1:]
    else
        return 0
    endif
    let link_infos = vimwiki#base#resolve_link(link)
    if link_infos.filename ==? ''
        echomsg 'Vimwiki Error: Unable to resolve link!'
        return 0
    else
        exe 'edit' . fnameescape(link_infos.filename)
        return 1
    endif
endfunction
```

rjshrjndrn:
It's working. Thank you.

tinmarino:
In `function! vimwiki#base#resolve_link(link_text, ...)` to many changes are made to the `link_text` even if this one is absolute (in which case it should not change).
I tried as clean as I could for easy integration : PR #755

justrajdeep:
I attempted to make `[//absolute_path]` modification to the plugin and created a PR#811.

tinmarino:
Fixed: #811 with 
```
* [Ansible refactor](///home/rajeshr/Documents/projects/public-devops/README.md)
```

PR #755, opened 2019-10-07
Fix: Markdown: point a file with absolute path #746

tinmarino:
Fixing absolute path links not working


tinmarino:
My bad, PR is bad : I forgot that https://github.com/vimwiki/vimwiki/blob/ba84981b5ae54af32e320cd04b583dd7e9fe1c6c/doc/vimwiki.txt#L1002-L1006

Issue #420, opened 2017-12-06
Add error handling to VimwikiSearch


glappen:
Hello, trying to get this setup the way I want and I have made 2 different wiki's:

```
let g:vimwiki_list = [{'path': '/home/greg/Nextcloud/vimwiki/Personal'}, {'path': '/home/greg/Nextcloud/vimwiki/LapComInc'}]
```

When I open a wiki and run :VimwikiSearch, I get this error:

```
"reg/Nextcloud/vimwiki/LapComInc/**/*.wiki" [New DIRECTORY]                                                                                                    
Cannot open file "reg/Nextcloud/vimwiki/LapComInc/**/*.wiki"                                                                                                   
E480: No match: home                                                                                                                                           
Press ENTER or type command to continue
```

The wikis seem to function normally otherwise.  Can anyway tell me where I went wrong here?

Thanks!!!

glappen:
I guess this was a really obscure error...is that the correct way to setup two different wikis at least?  Why is VimwikiSearch getting the path to the wiki files wrong?

Nudin:
Hi,
did you specify a search-string/pattern? Like `:VimwikiSearch foo`? If not – the error is not surprising: Vimwiki does not (jet) check is an argument is given – VimwikiSearch is simply bound to
`lvimgrep <args> yourpath/**/*.wiki`. If you don't give any argument, you get:
```
lvimgrep /home/greg/Nextcloud/vimwiki/LapComInc
```
So vim thinks `/home/g` is your search-pattern and searches for this in the path `reg/Nextcloud/vimwiki/LapComInc`.

One should probably handle the case when the user does not give an argument gracefully…

ranebrown:
The command should be changed to call a function with error handling built in.

ratfactor:
I'd like to contribute to VimWiki.  If no one else has started on this issue, it seems like a nice one for me to tackle.

ranebrown:
That would be great! Make sure to take a look at the test framework and include new tests as part of your PR. Let me know if you have any questions on that front.

ratfactor:
I have the fix completed.  Starting on tests.  I was very happy to see (and run) the test framework in the dev branch.  Good things are happening in this project.

ratfactor:
While I'm in here, I've noticed two other improvements to usability for VimwikiSearch:

### Unenclosed search patterns

If the user types a search pattern containing spaces without `/enclosing/` it, they'll get an error very similar to the original cause of this issue:

```
[...]

PR #757, opened 2019-10-12
Add error handling to VimwikiSearch

ratfactor:
The biggest challenge here was getting the Vader tests to work with the `lvimgrep` functionality which underlies the VimwikiSearch feature - after a lot of trial and error, the simplest and least fragile thing to do was make the `run_tests.sh` script run the Vader tests in separate Vim instances (but still a single Docker instance for the sake of efficiency) where they could not be "contaminated".

Commit message body:

 - Create function wrapper around lvimgrep for input checking, pattern
   quoting, and error handling.
 - Add Vader tests for VimwikiSearch.
 - Change syntax loading from try/catch to explicit file check (to
   prevent Vader test bug).
 - Update doc/vimwiki.txt for changes.
 - Change test script to run Vader tests separately

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1068,12 +1068,18 @@ function! vimwiki#base#nested_syntax(filetype, start, end, textSnipHl) abort
   " let b:skip_set_iskeyword = 1
   let is_keyword = &iskeyword
 
-  try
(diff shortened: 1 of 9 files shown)


ranebrown:
For #420

ranebrown:
Thanks!

Issue #691, opened 2019-04-26
VimwikiGoBackLink does not go back to links on the same page


hq6:
# Issue Checks

1. This issue reproduces on `dev` as well as master.
2. I am using the default syntax.
3. The vimrc is copied from the Troubleshooting page.
```
set nocompatible
filetype plugin on
syntax on
set runtimepath+=~/vw_tmp/vimwiki
let wiki = {}
let wiki.path = '~/vw_tmp/wiki'
let wiki.path_html = '~/vw_tmp/wiki/html'
let wiki.syntax = 'default'
let wiki.ext = '.wiki'
let g:vimwiki_auto_header = 1
let g:vimwiki_list = [wiki]
```
4. Output of `:VimwikiShowVersion`
```
Branch: dev
Revision: e84dcbf
Date: 2019-04-26 11:06:09 -0600
```

### Steps to Reproduce

1. Create an index page with the following contents: `[[AnotherPage]]`.
2. Follow the link and create the following contents
```
= AnotherPage =

== Header2 ==
```
3. Run `:VimwikiTOC` to generate the table of contents. The page now looks like this.
```
= Contents =
        - [[#AnotherPage|AnotherPage]]
                - [[#AnotherPage#Header2|Header2]]

= AnotherPage =

== Header2 ==



```
4. Put the cursor on line 3 over the link to `Header2`.
5. Press `<Enter>`. Observe that the cursor jumps to the header correctly.
6. Press `<Backspace>`. Observe that Vim returns to the wiki index page, instead of the TOC.

I think the correct behaviour of `<Backspace>` is to return the cursor to the location it came from, regardless of whether that location is in the same buffer or a different buffer.

ranebrown:
This seems to be as simple as changing [this line](https://github.com/vimwiki/vimwiki/blob/fd8832cfa1624267a8156b055de24c47ec966612/autoload/vimwiki/base.vim#L332-L333) to `let update_prev_link = is_wiki_link` but I didn't really look into the reasoning why links within the same file weren't being stored.

hq6:
@ranebrown , I've actually been investigating this. 
It's not that simple because there's only a single `prev_link` in each buffer.
Your suggestion has the desired affect but breaks a different functionality: being able to hit `Backspace` a second time and going back to the page from whence you came.

That is, you'd be able to go back to the link on the current buffer, but you have now lost the link to the page that jumped to the current buffer.
[...]

PR #766, opened 2019-10-24
Fix: VimwikiGoBackLink does not go back to links on the same page #691 (more stable pull request)

tinmarino:
Fixing issue #691

1/ `prev_link` -> `prev_links` which is not a link anymore but a list of links. So each buffer can have multiple prev_link
2/ Remove the check is_same_file before going back from link : `!vimwiki#path#is_equal(link_infos.filename, vimwiki#path#current_wiki_file())`

Stating #ranebrown: It does sound like storing multiple links would be required.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -104,6 +104,12 @@ function! vimwiki#base#find_wiki(path)
 endfunction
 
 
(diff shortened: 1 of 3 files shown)


ranebrown:
Can you try and add some Vader tests to cover your changes?

ranebrown:
Thanks!

Issue #747, opened 2019-08-30
Error in Documentation regarding new lines in MathJax lines 1272 & 1273


frrobert2:
The statement:
Note: no matter how many lines are used in the text file, the HTML will
compress it to one line only.
on lines 1272 and 1273 is incomplete.
MathJax will compress to one line unless you use \\ for a new line at the end of each line.
Example of 2 equations on their own lines within a block:
{{$%align%
x+3&=y\\
y&=x+5
}}$
will write each equation on its own line with the equal signs aligned.

ranebrown:
Statement in documentation [here](https://github.com/vimwiki/vimwiki/blob/ba84981b5ae54af32e320cd04b583dd7e9fe1c6c/doc/vimwiki.txt#L1379-L1380)

tessarin:
@frrobert2 I don't think this documentation is incorrect. That note refers to the `{{$...}}$` block which does always produce a one line equation. It is equivalent to LaTeX display math.

To produce an equation in more than one line, you do need an environment such as _align_. This is what you used in your example and is also what the documentation describes right after that note:

https://github.com/vimwiki/vimwiki/blob/ba84981b5ae54af32e320cd04b583dd7e9fe1c6c/doc/vimwiki.txt#L1382-L1389

tessarin:
@ranebrown While looking into this I noticed the instructions for remote loading MathJax don't work:

https://github.com/vimwiki/vimwiki/blob/ba84981b5ae54af32e320cd04b583dd7e9fe1c6c/doc/vimwiki.txt#L1414-L1417



The [official documentation](https://www.mathjax.org/#gettingstarted) recommends these instead:

```html
<script src="https://polyfill.io/v3/polyfill.min.js?features=es6"></script>
<script id="MathJax-script" async
        src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
```

Can I make a small PR to change that?

PR #768, opened 2019-10-25
Update MathJax CDN Loading Instructions

tessarin:
As mentioned in #747, loading MathJax from the CDN provided in instructions is outdated and may not work for some users.

This small PR updates those links to conform with the current instructions from MathJax (from its getting started [guide](https://www.mathjax.org/#gettingstarted)).


ranebrown:
Thanks!

Issue #774, opened 2019-11-11
[html conversion] Ignoring certain lines and pasting them to the html file as they are


kaphula:
Hey,

I was wondering how would it be possible to add custom `<div>` tags from the vimwiki file so that the tags would show up in the html conversion. With `<div>` tags,  I could decide case by case how certain elements are being displayed on the web page. As it is, I have to use uniform styling for all of the same kind of html elements.

Having said that, writing `<div>` tags straight into the vimwiki file kinda works, but the conversion system wraps them inside `<p>` tags. If it were possible to mark lines to be ignored during the conversion, this approach would probably work no problem as the `<div>` tags would appear in the html file without any modifications.

Unless there are methods in place that allows to do this already, I'd like to suggest a new feature that would work something like this. The name of the marker that triggers ignoring should probably be clean and short, but in this example it is  `<vimwiki_html_ignore>`

**How it looks in the vimwiki file:**
```
<vimwiki_html_ignore> <div id="move_to_left">
= My regular header =
My regular vimwiki paragraph.
<vimwiki_html_ignore> </div>
```

**How it looks in the html conversion:**

```
<div id="move_to_left">
<h1> My regular header </h1>
<p> My regular vimwiki paragraph </p>
</div>
```

Any tips on how to implement this feature are welcome, since I can also try to implement it myself, although I am not very experienced with vimscripting.

ranebrown:
I haven't done any work with the html conversion but you are welcome to take a look at it and submit a PR if you get something working.

kaphula:
Cool. I already got it to work with custom modification, but there's still an issue when "the line to be ignored" is not surrounded by empty lines, because otherwise it will be considered part of the preceding paragraph and the nested order of html elements break.

Here's the modification that enables line ignoring when the line starts with **%plainhtml** prefix.

This snipper can be inserted as the first "processed" check inside the function:
_autoload_ > _html.vim_ > _function! s:parse_line(line, state)_

```
  if !processed
    let trigger = '%plainhtml'
    if line =~# '^\s*' . trigger
      let pp = split(line, trigger)
      call extend(res_lines, pp)
      let processed = 1
    endif
  endif
```
If you know a quick fix for the issue described above or there is something fishy with the code snippet, please let me know. I literally started learning vimscript at the same time as I started to investigate how to add this feature. I'll do a pull request once I get it to work properly.

tinmarino:
Closed by your PR 778. Nice share @kaphula.

PR #778, opened 2019-11-24
Added new feature to insert plaintext to the final html conversion file.

kaphula:
Referenced issue: https://github.com/vimwiki/vimwiki/issues/774

As described in the issue, this pull request implements a feature to insert plaintext to the final html conversion file. This is useful if you want to apply custom css styling to specific parts of your vimwiki documents. 

Example usage inside vimwiki file:
```
%plainhtml <div style="text-align: center;">
My normal vimwiki paragraph that will be centered when viewed with html browser.
%plainhtml </div>
```
The feature triggers only when **%plainhtml** is the first thing on the line (excluding whitespace).

There was an issue when the **%plainhtml** was inserted on the same block with other elements such as paragraphs. The conversion script wrapped up the plaintext inside the corresponding html elements, creating incorrect nested html element structure:

```
%plainhtml </div>
My paragraph.
%plainhtml </div>
```
Was translated to:
```
<div>
<p> 
My paragraph.
</div>
</p>
```
Which is not what is wanted. The closing `</div>` should be the last element.

This was fixed by adding a list of checks that will close the conversion elements (table, para, quote...) before the plain text is inserted. If there are other items like these (or will be in the future), it might be necessary to add them to the list of checks. Now that I think about it, it might be a good idea to create a common function that does this checking for all of the conversion elements at one place.

I did not add any documentation yet since the keyword to trigger the plaintext (**%plainhtml**) may not be suitable. I learnt vimscript as I implemented this feature, so validating and testing the code further is also recommended.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1223,6 +1223,42 @@ function! s:parse_line(line, state)
 
   let processed = 0
 
+  if !processed
+    " allows insertion of plain text to the final html conversion
+    " for example:
+    " %plainhtml <div class="mycustomdiv">
+    " inserts that line to the final html file (without %plainhtml prefix)
+    let trigger = '%plainhtml'
+    if line =~# '^\s*' . trigger
+      let lines = []
+      let processed = 1
+
+      " if something precedes the plain text line,
+      " make sure everything gets closed properly
+      " before inserting plain text. this ensures that
+      " the plain text is not considered as
+      " part of the preceding structure
+      if processed && len(state.table)
+        let state.table = s:close_tag_table(state.table, lines, state.header_ids)
+      endif
+      if processed && state.deflist
+        let state.deflist = s:close_tag_def_list(state.deflist, lines)
+      endif
+      if processed && state.quote
+        let state.quote = s:close_tag_quote(state.quote, lines)
+      endif
+      if processed && state.para
+        let state.para = s:close_tag_para(state.para, lines)
+      endif
+
+      " remove the trigger prefix
+      let pp = split(line, trigger)[0]
+
+      call add(lines, pp)
+      call extend(res_lines, lines)
+    endif
+  endif
+
   " pres
   if !processed
     let [processed, lines, state.pre] = s:process_tag_pre(line, state.pre)
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -3466,6 +3466,7 @@ Contributors and their Github usernames in roughly chronological order:
     - Robin Lowe (@defau1t)
     - Abhinav Gupta (@abhinav)
     - Dave Gauer (@ratfactor)
+    - Valtteri Vallius (@kaphula)
 
 
 ==============================================================================


ranebrown:
I haven't had time to look at this but can you take a stab at adding some Vader tests?

kaphula:
I setup the testing environment and looked at the examples and the existing tests look somewhat reasonable, but what about with html conversion? I am not sure how to mix the Vader script's `Given` and `Expect` sections when `Given` works on .vimwiki file and `Expect` should work on the converted html file? Any tips on this? I am also on the #vimwiki IRC channel if you want to help me out sometime.

ranebrown:
You should be able to just open the converted HTML file with an `edit` command within vader and then confirm the contents are as expected.

kaphula:
Sorry, but I think testing the external html file output with vader is too complicated for me. I can't get the vader to convert my `wiki` files to `html` files and I don't know where the template of the default html page is for the test wiki, which is needed to setup a correct `Expect` section. I tried taking a look at `test/vimrc `for the path of the wiki and its html path, but the contents of `test/resources` and what is described in the` test/vimrc` do not match.

Here's a simple vader test I prototyped for starters:
```
Include: vader_includes/vader_setup.vader

Given vimwiki ():
    %plainhtml <div id="test">
    my paragraph
    %plainhtml </div>

Execute (Set syntax to default):
  :Vimwiki2HTML<CR>

edit ???

Expect (Markdown links):
    # the whole default html file should be here as a base + the modifications 
    # from "Given"
    ...
    <div id="test">
    <p>
    my paragraph
    </p>
    </div>
    ...

Include: vader_includes/vader_teardown.vader

```

tinmarino:
Merged with git and not github: 867d81a50cc (in dev branch)

1. Rebased your PR (just the tipical conflict in doc)
2. Add some test (f9b9b5a768)
3. Close issue #774
4. Close this PR.

Nice job. Feel free to make other PR, with tests it is much easier to accept.
By the way, very nice comments in your code which made review and mantenance easyer.

Here the test file
```vim
Include: vader_includes/vader_setup.vader

Execute (Copy Wiki's Resources):
  Log "Start: Copy Resources"
  call CopyResources()

Given (Void):

Execute (Edit TestHtml Wiki):
  edit $HOME/testwiki/TestHtml.wiki
  AssertEqual $HOME . '/testwiki/TestHtml.wiki', expand('%')
  AssertEqual 'default', vimwiki#vars#get_wikilocal('syntax')
  AssertEqual 0, vimwiki#vars#get_bufferlocal('wiki_nr')

Do (Markdwon with %plainhtml):
  :edit $HOME/testwiki/TestHtml.wiki\<CR>
  i%plainhtml<div id="test">\<CR>
  my paragraph\<CR>
  %plainhtml</div>\<CR>\<Esc>
  :set bt=\<CR>
  :write\<CR>

Execute (Save and Convert to html):
  edit $HOME/testwiki/TestHtml.wiki
  Vimwiki2HTML

Given (Void):

Do (Get Html body):
  :read $HOME/html/default/TestHtml.html\<CR>
# Goto body
  gg/<body>\<CR>
# Copy in b
  "bdat
# Delete All
  ggdG
# Paste body
  "bP
# Remove last line
  Gdd
# Save (Not necessary)
  :write

Expect (Plain Html):
# the whole default html file should be here as a base + the modifications 
# from "Given"
  <body>

  <div id="test">
  <p>
  my paragraph
  </p>
  </div>

  </body>

Include: vader_includes/vader_teardown.vader
```

PR #780, opened 2019-11-29
Suppress echoing filepath in custom_wiki2html call

rattletat:
Changes echomsg to call. This removes the "Press ENTER or type command to continue" message after each write when using a custom wiki2html script.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1437,7 +1437,7 @@ endfunction
 
 function! vimwiki#html#CustomWiki2HTML(path, wikifile, force)
   call vimwiki#path#mkdir(a:path)
-  echomsg system(vimwiki#vars#get_wikilocal('custom_wiki2html'). ' '.
+  let output = system(vimwiki#vars#get_wikilocal('custom_wiki2html'). ' '.
       \ a:force. ' '.
       \ vimwiki#vars#get_wikilocal('syntax'). ' '.
       \ strpart(vimwiki#vars#get_wikilocal('ext'), 1). ' '.
@@ -1454,6 +1454,10 @@ function! vimwiki#html#CustomWiki2HTML(path, wikifile, force)
       \     shellescape(s:root_path(vimwiki#vars#get_bufferlocal('subdir'))) : '-'). ' '.
       \ (len(vimwiki#vars#get_wikilocal('custom_wiki2html_args')) > 0 ?
       \     vimwiki#vars#get_wikilocal('custom_wiki2html_args') : '-'))
+  " Echo if non void
+  if output !~ "^\s*$"
+    echomsg output
+  endif
 endfunction
 
 


tinmarino:
@rattletat  thanks for your collaboration. I do prefer to see what my `Vimwiki2HTML` script is doing. So I can :
1. See errors
2. Debug my own script

If you don't like, you can map or call `:silent Vimwiki2HTML`. But if integrators accept your PR, I will not be able to see my output quickly.

What is your opinion?  Should the community share a `call` and read their output their way or `echomsg` and pipe their output to stdout, silent it whenever they want ?
Cheers.

rattletat:
Hi @tinmarino, 
ok, I was not aware of people using that feature, as I debugged my script in the terminal beforehand.

The thing bothering me is that this is the standard behavior when using the `auto_export` option, even if the custom script don't produce any output. Because in this case, you need to press twice when saving a file, which is quite inconvenient when live previewing your work. Could you test that please in your setup? Activating auto export and using a blank script.

autoload/vimwiki/html.vim
```diff
@@ -1437,7 +1437,7 @@ endfunction
 
 function! vimwiki#html#CustomWiki2HTML(path, wikifile, force)
   call vimwiki#path#mkdir(a:path)
-  echomsg system(vimwiki#vars#get_wikilocal('custom_wiki2html'). ' '.
```
tinmarino: Maybe echo it if not empty line

```vim
  let output = system(vimwiki#vars#get_wikilocal('custom_wiki2html'). ' '.
      \ a:force. ' '.
      \ vimwiki#vars#get_wikilocal('syntax'). ' '.
      \ strpart(vimwiki#vars#get_wikilocal('ext'), 1). ' '.
      \ shellescape(a:path). ' '.
      \ shellescape(a:wikifile). ' '.
      \ shellescape(s:default_CSS_full_name(a:path)). ' '.
      \ (len(vimwiki#vars#get_wikilocal('template_path')) > 1 ?
      \     shellescape(expand(vimwiki#vars#get_wikilocal('template_path'))) : '-'). ' '.
      \ (len(vimwiki#vars#get_wikilocal('template_default')) > 0 ?
      \     vimwiki#vars#get_wikilocal('template_default') : '-'). ' '.
      \ (len(vimwiki#vars#get_wikilocal('template_ext')) > 0 ?
      \     vimwiki#vars#get_wikilocal('template_ext') : '-'). ' '.
      \ (len(vimwiki#vars#get_bufferlocal('subdir')) > 0 ?
      \     shellescape(s:root_path(vimwiki#vars#get_bufferlocal('subdir'))) : '-'). ' '.
      \ (len(vimwiki#vars#get_wikilocal('custom_wiki2html_args')) > 0 ?
      \     vimwiki#vars#get_wikilocal('custom_wiki2html_args') : '-'))
  " Echo if non void
  if output !~ "^\s*$"
    echomsg output
  endif
```

tinmarino:
Hi @rattletat,

I reproduced your issue. And you are right. Well I did not know about `auto_export`.
Steps to reproduce :
```vim
let  vimwiki_wikilocal_vars[vimwiki#vars#get_bufferlocal('wiki_nr')]['custom_wiki2html'] = $HOME . "/void.sh"
let  vimwiki_wikilocal_vars[vimwiki#vars#get_bufferlocal('wiki_nr')]['auto_export'] = 1
echo vimwiki_wikilocal_vars[vimwiki#vars#get_bufferlocal('wiki_nr')]
```
Where `~/void.sh` is an empty file.
 
My Suggestion above satisfies both of us.
Unfortunately I did not find a way to remove the first message : `"test_wiki.md" 46L, 749C written` even with `set shortmess=filmnrwxaoOstTWAIcqF` did you ?

rattletat:
Hi @tinmarino,

thanks for your suggestion! It works for and does not produce any additional prompt for me, when running my output-free script.

Regarding the `"<filepath>" ..L, ..C` written message, I think this is standard vim behavior, and can only be avoided using some [remapping](https://www.reddit.com/r/vim/comments/5w4nns/disable_buffer_write_message_in_command_line/)

tinmarino:
`"test.txt" 42L, 69C` message. Thk ! You are right: just the `:silent w` or some silent call can silence that.
Can you integrate my changes in your PR ?

tinmarino:
@ranebrown  can you change " -> '   (cleaner)
* `if output !~ "^\s*$"` -> `if output !~ '^\s*$'`

ranebrown:
I'm in the process of updating everything to pass vint checks so this will be changed.

Issue #807, opened 2020-01-19
linking to existing diary pages results in only the first word of the page name being used


stevesteve:
Creating a link to a page in the diary path with multiple words in its title, from a page in the diary path, results in a link with only the first word of the page title.

To Reproduce:
* go to a diary page (e.g. `diary/2020-01-01.wiki`)
* write and, in visual mode, select text equal to an *existing* page, which is *in the diary path* and has *multiple words* in its name (e.g. `New Years Celebration`, if `diary/New Years Celebration.wiki` exists)
* press enter
* the text is replaced with a link, containing only the first word of the page name (e.g. "New") instead of the actual page name

VimwikiShowVersion:
```
Branch: dev
Revision: 75c557b
Date: 2020-01-08 06:03:40 -0700
```

PR #808, opened 2020-01-19
Fix for Issue #807

stevesteve:
creating a new link to an existing page in the diary path now uses the full page name instead of just the first word.

#807

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -2183,7 +2183,7 @@ function! vimwiki#base#normalize_link_in_diary(lnk) abort
     let template = vimwiki#vars#get_global('WikiLinkTemplate1')
   elseif link_exists_in_diary
     let str = a:lnk
-    let rxUrl = vimwiki#vars#get_global('rxWord')
+    let rxUrl = '.*'
     let rxDesc = ''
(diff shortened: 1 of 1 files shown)


autoload/vimwiki/base.vim
```diff
@@ -2183,7 +2183,7 @@ function! vimwiki#base#normalize_link_in_diary(lnk) abort
     let template = vimwiki#vars#get_global('WikiLinkTemplate1')
   elseif link_exists_in_diary
     let str = a:lnk
-    let rxUrl = vimwiki#vars#get_global('rxWord')
+    let rxUrl = '.*'
     let rxDesc = ''
     let template = vimwiki#vars#get_global('WikiLinkTemplate1')
```
tinmarino: That solves the bug.
But  better delete all the if block: it is the same as the latter else case there:
https://github.com/vimwiki/vimwiki/blob/75c557bcbca7d9ebf461629e2818e7c21786e6f7/autoload/vimwiki/base.vim#L2195-L2199

tinmarino:
OK: It solves the bug, and do not break anything on my machine. Nice Job !

Prettify: You should remove all 5 lines of the if clause: this removes redundant code.

Advice: Remove the spaces of your filename anyway

tinmarino:
Thank you @stevesteve

Issue #237, opened 2016-08-28
VimwikiRenameLink not working with Markdown


davidlmontgomery:
I find that VimwikiRenameLink doesn't work when using Markdown links like this 

`[Link text](link-file)
`

Because I am only using this kind of Markdown link (I am not using reference-style links), for my purposes the following hack worked to fix this issue. But it isn't a general solution.

```
--- vimwiki_markdown_custom.vim 2016-07-05 20:17:09.084795600 -0700
+++ vimwiki_markdown_custom.vim 2016-07-05 20:16:22.290611000 -0700
@@ -31,8 +31,8 @@
 " 1. [URL][], or [DESCRIPTION][URL]

 let s:wikilink_md_prefix = '['
-let s:wikilink_md_suffix = ']'
-let s:wikilink_md_separator = ']['
+let s:wikilink_md_suffix = ')'
+let s:wikilink_md_separator = ']('
 let s:rx_wikilink_md_prefix = vimwiki#u#escape(s:wikilink_md_prefix)
 let s:rx_wikilink_md_suffix = vimwiki#u#escape(s:wikilink_md_suffix)
 let s:rx_wikilink_md_separator = vimwiki#u#escape(s:wikilink_md_separator)

```

EinfachToll:
Depends on #256

ranebrown:
Renaming a page should now work but the links are not correctly updated. Related to #406

ranebrown:
@davidlmontgomery This should be fixed on the latest `dev` branch.

f-dj:
> @davidlmontgomery This should be fixed on the latest `dev` branch.
Can confirm that it works in most cases, but still does not work with whitespaces. Is this expected behaviour?

I mean, I was surprised that whitespaces in links work in the first place. But shouldn't the renaming work also for consistency if you are allowed to generate them automatically (by going to visual mode, select multiple words, hit enter)? Feels like a minor bug.

f-dj:
I opened PR #824 to make the `find_files` function work for filenames with whitespaces. However, on master the `split` method already took `'\n'` as  2nd parameter. So it should only affect the dev branch. 

Would be good to merge this PR anyway IMO, so that the dev branch is as functional as master. Also, this prevents a bug from being introduced if dev is eventually merged into master.

etiennekintzler:
The issue still appear in my case (file is renamed but links are not updated). 

This is the case _whether there is a space or not_ in the filename. 

[...]

PR #824, opened 2020-02-24
fix base#find_files so that it works for filenames with whitespaces

f-dj:

#237
  - Fix base#find_files so that it honors whitespaces in file names. The problematic code could only be found on the `dev` branch. The fix for #237 on `dev` does not work if a whitespace is in the filename. As long as this feature is not on `master`, this PR makes it work on the `dev` branch. 


f-dj:
Reopened as pull request #828 with dedicated branch, as other commits found their way into this one.

PR #828, opened 2020-03-11
fix base#find_files so that it works for filenames with whitespaces

bratekarate:

Clone of #824, other commits accidentally found their way into the old one.

#237
  - Fix base#find_files so that it honors whitespaces in file names. The problematic code could only be found on the `dev` branch. The fix for #237 on `dev` does not work if a whitespace is in the filename. As long as this feature is not on `master`, this PR makes it work on the `dev` branch. 
--> probably not acclipable

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -498,11 +498,11 @@ function! vimwiki#base#find_files(wiki_nr, directories_only) abort
   else
     let pattern = '**/*'.ext
   endif
-  let files = split(globpath(root_directory, pattern))
+  let files = split(globpath(root_directory, pattern), '\n')
 
   " filter excluded files before returning
   for pattern in vimwiki#vars#get_wikilocal('exclude_files')
-    let efiles = split(globpath(root_directory, pattern))
+    let efiles = split(globpath(root_directory, pattern), '\n')
     let files = filter(files, 'index(efiles, v:val) == -1')
   endfor
 


tinmarino:
I did not try but it seems more coherent than a split on whitespace.
In the dev branch, `VimwikiRenameLink` works with file with spaces.

But for nitpick. Now file cannot have a newline inside :-)
From `:h globpath` -> better use a list to be filled in the argument
```text
		When {list} is present and it is |TRUE| the result is a List
		with all matching files. The advantage of using a List is, you
		also get filenames containing newlines correctly. Otherwise
		the result is a String and when there are several matches,
		they are separated by <NL> characters.  Example: >
```

tinmarino:
Well I've just tested (after rebase) and it does not break anything. It must fix some stuff, but I don't know what.
Anyway. It is more robust that way, that you.

bratekarate:
> In the dev branch, `VimwikiRenameLink` works with file with spaces.

`VimwikiRenameLink` did not work for me for files with spaces on `dev`. Maybe this has been fixed by some other change, i don't know. But until now, I used my own fork with the newline change in `base#find_files` because I had issues with `dev`. But now it is merged anyway :smiley: 

> But for nitpick. Now file cannot have a newline inside :-)
> From `:h globpath` -> better use a list to be filled in the argument
> 
> ```
> 		When {list} is present and it is |TRUE| the result is a List
> 		with all matching files. The advantage of using a List is, you
> 		also get filenames containing newlines correctly. Otherwise
> 		the result is a String and when there are several matches,
> 		they are separated by <NL> characters.  Example: >
> ```

It seems you are right about that. Not really an expert in vimscript. I used trial and error first, then saw that `split` with `\n` is used on `master` as well. This was enough for me to assume that it's not too hacky.

Thanks for merging, now I can use `dev` locally. :)

Issue #831, opened 2020-03-12
Feature Request: Support GFM anchor links inside wiki files


bratekarate:
Prior to submitting a new issue make sure to complete these steps:

- [x] Checkout the `dev` branch and confirm the issue is present there as well.
      The `dev` branch contains fixes that may not have been merged to `master` yet.
- [x] Post the syntax you are using (default/mediawiki/markdown) **and** your vimwiki settings from your `.vimrc`
- [x] Provide a detailed description of the problem including **steps to reproduce the issue**.
- [x] Include the output of `:VimwikiShowVersion`.

- Syntax used: markdown
## Description
Links to Headers/Anchors require as url the exact same text as the header that is linked to. Otherwise they don't work. Example:
```
[Link to heading](#Heading Example)
## Heading Example
```
For HTML export, e.g. with pandoc, or when using the same wiki content for an engine that uses GFM, it is desirable to use hyphens instead of spaces in the link url and use only lowercase. For a table of contents that is usable after HTML export, plugins like [mzlogin/vim-markdown-toc](https://github.com/mzlogin/vim-markdown-toc) can be used. Example:
```
[Link to heading](#heading-example)
## Heading Example
```
This works great for export, but inside vim the links are not usable. It would be great if vimwiki would also recognize these links, for instance by first trying the exact name and afterwards with lowercase and hyphens.

ranebrown:
This is something I've been wanting to fix as well, just haven't had the time. If you want to open a PR that would be great.

bratekarate:
If I find the time I may be able to. Haven't checked out the responsible code yet though and don't know how complex this change would be.

bratekarate:
Done, #840. Not sure if that's the way to go, but it works. Tell me if it should be improved, don't know if I find any more time soon. 

Also (although unrelated), #828 is still open and ready to merge.

bratekarate:
@ranebrown have you checked out my PR yet?

ranebrown:
This is actually a duplicate of #666. [This comment](https://github.com/vimwiki/vimwiki/issues/666#issuecomment-570753457) has some useful info.

tinmarino:
Close with PR #840 in favor of Issue #666 trakcing the remaining support

folofjc:
Has this been pushed to `master`? This functionality doesn't appear to work for me and I cannot find it in the commit history of `master`.

Issue #666, opened 2019-04-11
Anchor support


ElModdy:
Is there a way to set an anchor like the following one?
https://gist.github.com/asabaylus/3071099
I appreciate any kind of help

ranebrown:
Vimwiki supports anchors as noted in the help file `vimwiki-anchors`. For markdown syntax they will be formatted as shown below when using `:VimwikiTOC`
```markdown
# Contents

- [Header1](#Header1)
    - [Header2](#Header1#Header2)
- [Header 1 Spaces](#Header 1 Spaces)
    - [Header 2 Spaces](#Header 1 Spaces#Header 2 Spaces)

# Header1

## Header2

# Header 1 Spaces

## Header 2 Spaces
```

This format isn't compatible with GitHub and GitLab which downcase, drop special characters, replaces spaces with `-` and add a number if necessary to make the link unique. I think it would make sense for Vimwiki to follow this format when using Markdown syntax but it isn't currently supported. If you want to open a PR to modify this behavior that would be great!

ElModdy:
Thanks a lot for the replay, one more quest can I make an anchor that isn't a header?

ElModdy:
Solved just looking at  the doc

ranebrown:
I'm reopening this as an enhancement. I would eventually like to transition to GitHub style anchor links.

ranebrown:
> The code that creates the anchors is here:
> https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb
> 
>     1. It downcases the string
> 
>     2. remove anything that is not a letter, number, space or hyphen (see the source for how Unicode is handled)
> 
>     3. changes any space to a hyphen.
> 
>     4. If that is not unique, add "-1", "-2", "-3",... to make it unique

tinmarino:
Thanks @ranebrown for the [list](https://github.com/vimwiki/vimwiki/issues/666#issuecomment-570753457), From (PR #840) (Issue #831) 1 and 3 are done

Still missing:
- [ ] 2. remove anything that is not a letter, number, space or hyphen (see the source for how Unicode is handled)
- [ ] 4. If that is not unique, add "-1", "-2", "-3",... to make it unique
- [ ] Link generation: #840 only consider link movements

tinmarino:
Fixed in dev: (428c60a: or Search multiple heading with same name)

Related: #968

Issue #624, opened 2019-03-08
More maintainers for vimwiki


Nudin:
Vimwiki currently only has very little maintainers. PRs are waiting very long until they get merged or commented. This causes frustration and delays the project.

New maintainers are appointed by @EinfachToll. Use this Issue if you want to apply for maintainerrights.

This issue superseeds the older issue #402.

ranebrown:
I'm interested in being a maintainer. Let me know if you need any other info. Thanks!

hq6:
I am also interested in being a maintainer / developer. I have some experience with `vimscript` and am excited about adding new features. Additionally, I have enough reputation on StackOverflow to create a tag there and redirect questions to that community is it is desirable.

EinfachToll:
I will add you both as participants. I also added a few advice to CONTRIBUTING.md for you. Hopefully you are aware of your huge responsibility and not only try to add new fancy features but fix old and boring bugs ;)

P.S. Just so you know, when I'm sometimes inactive for a few weeks, that doesn't mean anything. My contributions come slowly, but I don't see me stopping or growing tired or something.

EinfachToll:
OK, I think I've sent you invitations to be collaborators. Let me know if you need any rights you don't have. I don't completely understand the rights management here on Github.

hq6:
@EinfachToll, I do not find `CONTRIBUTING.md` in the repository. Can you offer a link?

Nudin:
@hq6 it's in the [dev-branch](https://github.com/vimwiki/vimwiki/blob/dev/CONTRIBUTING.md)!

Nudin:
@EinfachToll could you please add me to the `vimwiki` Github-group? So that I can create new repositories. (Beside others for further documentation and testing)

hq6:
If we're going to have separate but related repositories, it may be more
efficient if you add me and Rane to well.

Otherwise, we'd have to be manually added as collaborators on each repo
after creation.

thecliguy:
Just in case any maintainers don't know, there is a vimwiki IRC channel. I'd encourage you to join, details below:

`#vimwiki` on Freenode ([webchat](https://webchat.freenode.net/?channels=#vimwiki), also synced to Matrix/Riot: `#freenode_#vimwiki:matrix.org`)

brennen:
Howdy all - I'd be happy to at least help triage issues and such from time to time...

tinmarino:
[...]

PR #840, opened 2020-03-25
Allow hyphens instead of spaces and search case-insensitive in jump_to_anchor function

bratekarate:
**Disclaimer: Log is kind of messed up because I maintain my own fork and needed to pull the latest version in to my repo** 
 
This PR allows GFM style links to be used (Issue #831). `jump_to_anchor` has been altered so that search for anchors is now case-insensitive and dashes may be included instead of whitespaces.

Code works as follows: before the anchor is transformed into the regex patterns, a `\c` is prepended before the first character of each word and spaces are replaced by `[ -]`. This way the search function also matches GFM style links.

I'm not a vimscript expert, so I don't know if this solution is the prettiest, but it is a one liner. Feel free to improve if I missed something or if it's too hacky.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -659,6 +659,10 @@ function! s:jump_to_anchor(anchor) abort
 
   for segment in segments
 
+    " prepare segment pattern so that it is case insensitive and also matches dashes
+    " in anchor link with spaces in heading
+    let segment = substitute(substitute(segment, '\<\(.\)', '\\c\1', 'g'), '-', '[ -]', 'g')
+
     let anchor_header = s:safesubstitute(
           \ vimwiki#vars#get_syntaxlocal('header_match'),
           \ '__Header__', segment, '')
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -659,9 +659,12 @@ function! s:jump_to_anchor(anchor) abort
 
   for segment in segments
 
-    " prepare segment pattern so that it is case insensitive and also matches dashes
+    " Craft segment pattern so that it is case insensitive and also matches dashes
     " in anchor link with spaces in heading
-    let segment = substitute(substitute(segment, '\<\(.\)', '\\c\1', 'g'), '-', '[ -]', 'g')
+    " Ignore case
+    let segment = substitute(segment, '\<\(.\)', '\\c\1', 'g')
+    " Treat - as [- or space]
+    let segment = substitute(segment , '-', '[ -]', 'g')
 
     let anchor_header = s:safesubstitute(
           \ vimwiki#vars#get_syntaxlocal('header_match'),


bratekarate:
Fixed messed up log.

ranebrown:
Although this implementation partially works I think there are some cases that might be missed. See [this comment](https://github.com/vimwiki/vimwiki/issues/666#issuecomment-570753457) of #666.

Things like `VimwikiTOC` and tags within two lines of a header for `VimwikiGenerateTagLinks` should be supported as well.

bratekarate:
I'm not sure I understand why `VimwikiTOC` has snomething to do with this.

The Issue  was not intended to be about the `VimwikiTOC` generating GFM compatible markdown, but rather to make vimwiki compatible with GFM style anchors. This implementation works with both `VimwikiTOC` anchors and GFM style anchors. At least I haven't found a `VimwikiTOC` generated anchor that did not work for me.

Are you saying the generation with `VimwikiTOC` should be modified in the same issue? Allowing vimwiki to recognize GFM style anchors -- what I did here -- and generating a valid GFM style TOC sounds like two different issues to me.

As for tags within two lines of a header, I am not too familiar with this functionality yet. Do you mean that tags in within two lines of a header can be referenced by anchors?

ranebrown:
> Are you saying the generation with VimwikiTOC should be modified in the same issue?

It would be nice if auto generated links to headers used the GFM format but I'm ok with that being a separate issue if you don't want to add it to this one.

The bigger concern is making sure all headers are properly supported such as ones with special characters.

> Numberings for duplicates could be ignored to always jump to the first occurence.

Sounds reasonable.

bratekarate:
> It would be nice if auto generated links to headers used the GFM format but I'm ok with that being a separate issue if you don't want to add it to this one.

I'm afraid I don't have the time right now. I may be able in the future, but I am not sure yet. Hence I would like to keep this a separate issue that can be resolved soon. This way it is possible to e.g. copy existing GFM style markdown in vimwiki and navigate the anchors.

> The bigger concern is making sure all headers are properly supported such as ones with special characters.

That's clear. I will improve the code. 

I just tried using links to tags within 2 lines of a heading with `VimwikiGenerateTagLinks`, it still works with this PR. Of course only within vim, not in GFM Markdown. This would have to be reworked together with `VimwikiTOC`

ranebrown:
Probably makes sense to create a function in `autoload/vimwiki/u.vim` to handle the link transformation.

YangVincent:
How can I get this working in the meantime? I've made the same change in my autoload/vimwiki/base.vim, but it doesn't seem to trigger on "enter".

YangVincent:
Update -- I ended up adding 
`let fname = tolower(substitute(fname, ' ', '-', 'g'))` on line 845 in vimwiki#base#edit_file, in `base.vim`.

tinmarino:
> `let fname = tolower(substitute(fname, ' ', '-', 'g'))` on line 845 in vimwiki#base#edit_file, in `base.vim`.

This is very bad. You got cannot jump to files with Uppercase or spaces

>     let segment = substitute(substitute(segment, '\<\(.\)', '\\c\1', 'g'), '-', '[ -]', 'g')

This seems a possible start.
A little hacky really ? Much too much for me ! See in the PR.

autoload/vimwiki/base.vim
```diff
@@ -659,6 +659,10 @@ function! s:jump_to_anchor(anchor) abort
 
   for segment in segments
 
+    " prepare segment pattern so that it is case insensitive and also matches dashes
+    " in anchor link with spaces in heading
+    let segment = substitute(substitute(segment, '\<\(.\)', '\\c\1', 'g'), '-', '[ -]', 'g')
```
tinmarino, line 664: Why one liner ?
```
" Craft Segment regexp
" Ignore case
let segment = substitute(segment, '\<\(.\)', '\\c\1', 'g')
" Treat - as [- or space]
let segment = substitute(segment , '-', '[ -]', 'g')
```

tinmarino:
From [the famous comment](https://github.com/vimwiki/vimwiki/issues/666#issuecomment-570753457): still missing the opposite of:

2. remove anything that is not a letter, number, space or hyphen (see the source for how Unicode is handled)
4. If that is not unique, add "-1", "-2", "-3",... to make it unique

tinmarino:
So what must be done before using the anchor:

* [ ] Change the one liner way to something more readable (please) (5 min)
* [ ] Add the letter, number, space or hyphen potential insertion anywhere (point 2) (15 min)
* [ ] Add the -1 potential removal (point 4) (5 min)
* [ ] Add some non regression tests (2h for the first time, 15 min afterwhile)

@bratekarate can you do that ? I can take the non regression test if you dont want to install the docker

tinmarino:
Closed with (PR #840)  in favor of (Issue #666) to track what is left

tinmarino:
FIxed: 355c1f7 
I made the modification (unhack the hacky) line and added test in the commit referencing this issue on dev 

@bratekarate  Thank you very much for the collaboration.
Feel free to make other small commits like that, if you describe the tests as well as in  (Issue #831), I'll write the tests and integrate faster.

Some related changes can be found at #666

bratekarate:
@tinmarino Thank you so much for for completing this fix! No need to thank me, I punched in a one liner and then pretty much ghosted this PR whilst saying I would continue to work on it. Your test cases will probably help me better understand the testing framework in use.

I'm currently in the final stage of my studies at university and can't find any time for github projects in my free time. That is also why I did not respond. I look forward to further contributing to vimwiki after I finally finish my studies, as I still use vimwiki daily at work and privately. Without it, I probably would have to switch to Doom emacs and use org mode.

Also, sorry for necrobumping this PR with non-technical boilerplate text :)

tinmarino:
Hi @bratekarate,

Happy to see some people with the will to work. Don't worry, the project can wait for you. Once you feel ready: you have some time and want to code, better than making flappy bird, try to fix one or two [good first issues](https://github.com/vimwiki/vimwiki/issues?q=is%3Aissue+is%3Aopen++label%3A%22good+first+issue%22+).
It is now a Github convention to use this label and leave some "not so hard" stuff for easy joining. We would be very glad if you could join our ranks (See #624 for the entry door I used 3 month ago).

> Without it, I probably would have to switch to Doom emacs and use org mode.

Same here. Plus Emacs has the advantage that it is not What You See is What You Get but the big drawback that it is never on my machine

> Also, sorry for necrobumping this PR with non-technical boilerplate text :)
 
I forgive you, if you forgive me  🐻. In reality, it is always motivating to see some happy users and much more to treat with some potential contributors (Imagine if would be 10 for 2 months on Vimwiki: even Emacs people would come)

bratekarate:
Glad to hear that there are "not so hard" things to fix, since that probably fits my almost nonexistent vimscript skills. I will check out some of these said labelled issues as soon as I find time. :+1:

PR #889, opened 2020-05-22
 Allow list symbols to be configured per wiki

Nudin:
Yet another attempt to finally get this fixed and upstream…
See also #390 and #479 for earlier attempts


Nudin:
We should merge this only after the 2.5 release.

Nudin:
This should now be finally finished and can be merged after #895 is merged. The failing tests are due to the bug fixed in #895.

Nudin:
I consider this ready to merge into `dev`. What do you think, @tinmarino @ranebrown?

tinmarino:
In my opinion, merge ASAP: the dev branch is here to collaborate so avoid diverging.
Plus I'm adding test, so I may change your test or see bugs sooner

Issue #894, opened 2020-05-30
Make all appropriate settings wikilocal


Nudin:
For historic reason some settings are global even though there is no reason to have them not as wikilocal settings. Therefore, it isn't possible to have multiple wikis with different values for these settings.

These are settings that affect the wikitext and should therefore be changeable on a per-wiki scope:
- [x] vimwiki_listsyms
- [x] vimwiki_listsym_rejected
- [x] vimwiki_list_ignore_newline (5700313e882814)
- [x] vimwiki_text_ignore_newline
- [x] vimwiki_markdown_link_ext (4672deb44e34b8)
- [x] vimwiki_toc_header, vimwiki_toc_header_level, vimwiki_toc_link_format (c4570886b04)

These are options that only change the displaying, they could be wikilocal but for most users it's also fine if they are global only:

- [x] vimwiki_hl_headers
- [x] vimwiki_hl_cb_checked
- [x] vimwiki_folding
- [x] vimwiki_create_link

The following options only make sense as global options and should stay such:
- vimwiki_use_calendar
- vimwiki_menu
- vimwiki_ext2syntax
- vimwiki_global_ext

Related Issues:

- [X] Unify global and wiki-local user options #257  (Closed in favor of this one)
- [x] make `g:vimwiki_toc_header` a per-wiki option #127
- [x] Support for wikis without adding path to g:vimwiki_list #854
- [x] Pressing enter to create a [[double bracket]] #892
- [x] Allow setting default path_html for temporary wikis #250

Nudin:
PR #889 makes the first two (`vimwiki_listsyms` & `vimwiki_listsym_rejected`) wiki-local. To do so it adds "tooling" to have settings that are usable as global and wiki-local. Through this vimwiki stays backward compatible: global settings continue to work but local options are used if pressent.

tinmarino:
Hacktoberfest point for any issue referenced or referencing this one and some are pretty easy: just dictionary refactoring.

tinmarino:
I think most of the refactoring have been done by now. If someone have specific variables to move, please link the new issue to this one.

Issue #892, opened 2020-05-29
Pressing enter to create a [[double bracket]]


santiyounger:
Hey,

I'm new to github, apologies if this is not properly structure, if you need more info let me konw. 
**This is probably an easy solve but I am stuck trying to solve it. Any help or suggestions are greatly appreciated.**
I'm on vimwiki Stable Version 2.5.
I recently moved my whole set up from ubuntu to manjaro kde plasma. Vimwiki is my post important vim plug-in and I absolutely love it.

**I rebuilt my vimrc with all the plug-ins. vimwiki seems to work alright, the only problem I have is the ENTER fucnction.**

When I press enter on some text to create a new vimwiki markdown file. I get this kind of situation.

**[new note](new note)**

I used to have this kind of result

**[[new note]]**

I really want to go back to the double bracket and I don't know why that changed since everything in my vimrc is the same.

I istanlled vim wiki with Plug like this:

`Plug 'vimwiki/vimwiki'`

Here I'll copy everything I have in my vimrc regarding vimwiki.

Vimrc
```
". .............................................................................
  Vim wiki
"............................................................................

"set nocompatible
filetype plugin on
syntax on

"".... vimwiki with markdown support
"let g:vimwiki_ext2syntax = {'.md': 'markdown', '.markdown': 'markdown', '.mdown': 'markdown'}
"".... helppage -> :h vimwiki-syntax

let g:vimwiki_list = [{'path': '~/vimwiki/',
                      \ 'syntax': 'markdown', 'ext': '.md'}]

```

sharingtom:
Hi, I just want to chime in and say that I would also very much prefer the double brackets in Markdown files.

If Markdown style links are the default for the Markdown syntax, would you consider adding something like _g:vimwiki_link_format_, or an option to _g:vimwiki_list_?

Nudin:
> ****
> 
> I really want to go back to the double bracket and I don't know why that changed since everything in my vimrc is the same.

We just released v2.5 a few days ago, so you likely have upgraded vimwiki without realizing when moving to the new setup.
I am however a bit confused, I tried with 2.3, 2.4.1 and 2.5 and in all three cased the markdown syntax `[text](text)` is used.

@ranebrown @tinmarino: Do you know if there was a change in 2.5 that affects this?

@sharingtom: Do you also have a regression (meaning previously double brackets in Markdown worked) or is this "just" a feature request?

@santiyounger: Can you try downgrading to 2.4 and check if it works then again?
[...]

PR #895, opened 2020-05-30
Bugfix: file extension for markdown is syntax wide

Nudin:
`populate_extra_markdown_vars()` contained uses of the file-extension via
`vimwiki#vars#get_wikilocal`. Since the first is syntax-wide and the second is
wikilocal and no wiki is specified or open at the moment of the call, the file
extension of the first markdown wiki will be used for all markdown wikis.

This PR fixes this via use of the existing template infrastructure.

This bug becomes critical when merging #889 or fixing #894.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1023,8 +1023,8 @@ function! s:update_wiki_links(wiki_nr, dir, old_url, new_url) abort
     let old_url_r = '\%(\.[/\\]\)\?' . old_url_r
     " Compute old url regex with filename between \zs and \ze
     let old_url_r = vimwiki#base#apply_template(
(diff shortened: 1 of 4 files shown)


Nudin:
The alternative would have been moving `WikiLinkMatchUrlTemplate` to the wiki-local settings altogether, but since we already have and use the template filling function I thought this is the cleaner approach. Issue #892 might however make it nessecary to move it nethertheless?

tinmarino:
> The alternative solution: moving `WikiLinkMatchUrlTemplate` to the wiki-local settings altogether

Yes good idea, but currently a premature optimization.

This PR is working towards separating __wiki__ and __syntax__ which is the good way. Once they are separated, it will be really easy to:
1. Load only the syntax we need
2. Even make 3 autoload files
3. Make an user accessible syntax dic
4. Add #892 feature request

This way, we can move the syntax feature requests to the user scope.

PR #391, opened 2017-09-16
WIP: Clean tasks that are done

Nudin:
Adds functions to clean up you todo-lists by removing all items marked as done.

PR #906, opened 2020-06-05
Functions to remove all todo items that are done

### LINKED ISSUE #823: Archive todo lists
Would be nice to have for Vimwiki, was using it back when i was on Sublime text with [PlainTask](https://packagecontrol.io/packages/PlainTasks) they had it so you could put the archive under the list or in a separate file, really great way to keep track on what has been done and to keep your current list from getting to big with things you already completed.

up

Maybe i should get another plugin for task management but i thought it would be a good ide for Vimwiki

Nudin:
This adds functions that remove todo items that are marked as done.

* `VimwikiRemoveDone`/`vimwiki#lst#remove_done` removes them within the selected area.
* `VimwikiRemoveDoneInList`/`vimwiki#lst#remove_done_in_list` removes them in the list the cursor is in.
  * `vimwiki#lst#remove_done_in_list` has a parameter to define if it should work recursively. Default is not recursive.

Superseeding #391
Fixes #823 and others.

### DIFF
--- a/autoload/vimwiki/lst.vim
+++ b/autoload/vimwiki/lst.vim
@@ -1157,7 +1157,8 @@ function! vimwiki#lst#remove_done(first_line, last_line)
 
   let parent_items_of_lines = []
   let cur_ln = first_item.lnum
-  while cur_ln > 0 && cur_ln <= last_item.lnum
+  let end_ln = last_item.lnum
+  while cur_ln > 0 && cur_ln <= end_ln
     let cur_item = s:get_item(cur_ln)
     if s:is_done(cur_item)
       let cur_parent_item = s:get_parent(cur_item)
@@ -1165,6 +1166,8 @@ function! vimwiki#lst#remove_done(first_line, last_line)
         call insert(parent_items_of_lines, cur_parent_item)
       endif
       exe cur_ln.'delete _'
+      let cur_ln -= 1
+      let end_ln -= 1
     endif
     let cur_ln = s:get_next_line(cur_ln)
   endwhile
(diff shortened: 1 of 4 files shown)


Nudin:
The documentation still has to be improved. I wasn't sure if we should define both commands and whether default should be recursive or not. Recursive has the somewhat surprising but correct side effect that the status of items can decrease. See the test for example.

@tinmarino @rane-brown any thoughts?

tinmarino:
I would put recursive as default: to get most work done which is harder to do by hand.

Very readble tests, really awesome.
Reading the last one, with spaces suprised me it was considered as a single list but better. Can you add tests for the behavior of 2 lists with content in between ?

```
* [X] Done 1
* [ ] Todo 1

This should break the list

* [ ] Todo Post space
* [X] Done Post space
```

ubitux:
It probably doesn't belong in this PR, but what about multi-line TODO entries? I often use `gq` and ends up with multi-line entries, and I know they're not well supported in vimwiki (at least in markdown `let g:vimwiki_hl_cb_checked = 1` will only gray-out the first line). So I'm curious how the removal of this PR behaves here in that situation.

Nudin:
> I would put recursive as default: to get most work done which is harder to do by hand.

Good argument, I changed it to recursive and updated the tests accordingly.

> Reading the last one, with spaces suprised me it was considered as a single list but better. Can you add tests for the behavior of 2 lists with content in between ?

I actually found out about this behavior by your `list_update.vader` tests. ;) This PR doesn't change anything on the general behavior of list but reuses existing functions.

> It probably doesn't belong in this PR, but what about multi-line TODO entries? I often use `gq` and ends up with multi-line entries, and I know they're not well supported in vimwiki (at least in markdown `let g:vimwiki_hl_cb_checked = 1` will only gray-out the first line). So I'm curious how the removal of this PR behaves here in that situation.

Yes, better create a ticket for that (if it doesn't exist yet).

Issue #942, opened 2020-07-20
HTML conversion on Windows fails to correctly convert some wiki links


adi93:
Vimwiki2HTML converts some links wrongly on Windows. I have tested this on Windows 10
TL;DR: Windows backslashes are causing a problem.
# Setup

Output of `:VimwikiShowVersion` command:

    Os: Windows
    Vim: 802
    Branch: dev

    Revision: da9ef92

    Date: 2020-06-16 23:21:41 -0400

Syntax I am using: **default**

Current vimwiki settings:

    let wiki_1 = {}
    let wiki_1.path = '~/wiki/'
    let wiki_1.path_html = '~/wiki/html/'
    let wiki_1.index = 'index'
    let wiki_1.template_path = '~/notes/templates/'
    let wiki_1.template_default = 'default'
    let wiki_1.template_ext= '.html'

    let g:vimwiki_list = [wiki_1]
    let g:vimwiki_table_mappings = 0

File structure: I have the following files on Windows (On Windows, ~ translates to `C:\Users\user`) :
    
    C:\Users\user\wiki\index.wiki
    C:\Users\user\wiki\foo\bar\abc.wiki

Contents of `index.wiki` file:

    = Home =
    [[./foo/bar/abc|Link to abc file]]

Contents of `abc.wiki` file:

    = ABC =
    [[../../index|Link to home]]
# Steps to reproduce
Open the abc.wiki file and run `:Vimwiki2HTML`. 
# Bug
After conversion, the "Link to home" link in `abc.html` should point to `C:\Users\user\wiki\html\index.html`. Instead, it points to `C:\Users\user\wiki\html\foo\bar\index.html`, which does not exist.

## Reason for bug

1. The `wikifile` variable passed to the `convert_file` function in `html.vim` is being passed as `C:\Users\user\wiki\foo\bar\abc.wiki` (line 1646 in `html.vim`.), which sets the global variable `current_wiki_file`.
2. This is then passed to the `vimwiki#base#resolve_link` function (line 482 in `html.vim`).
3. The `resolve_link` function then takes the root of source, and then forms the link for `index.wiki` as `C:\Users\user\wiki\foo\bar/../../index.wiki`, which is then passed to `vimwiki#path#normalize` function (line 267 in `base.vim`),
4. which returns `C:\Users\user\wiki\foo\bar/index.wiki`, instead of the expected `C:\Users\user\wiki\index.wiki`

# Fix
On line 1646 in html.vim, in the function , we have:

[...]

PR #945, opened 2020-07-20
vimwiki#942: Fixing link resolution in windows when converting to html

adi93:
Ref:#942

Fixing wrong html link conversion in windows, sanitizing path with `wikify_path` before calling convert_file.


tinmarino:
Good job.
I'll wait to merge so that I remember to make some non regression tests (copying your detailled issue).
It seems perfect anyway.

PR #946, opened 2020-07-20
Add option g:vimwiki_commentstring to disable vimwiki commentstring

BirgerNi:

By default the commentstring ist set to `%%%s`. For markdown documents which should be exported to html via pandoc a commentstring like `<!--\ %s\ -->` may be more suitable. This PR provides the option `g:vimwiki_commentstring` to turn off vimwiki commentstring. The option ist turned on by default to mainain backwards compatibility.


tinmarino:
Nice, I did not know this feature.

So you `let g:vimwiki_commentstring = 0`  __and__ `set commentstring='<!-- %s -->`
You think about what I think ? Why don't you just `let g:vimwiki_commentstring = '<!-- %s -->'`  and null string to prevent vimwiki setting the vim var and unset to keep default behavior (which I dont like by the way).
In other words, let the option give the chance for the user to add some customization.
Would you mind adding some 3 regression tests ?

All that said, thank you for the contribution.

BirgerNi:
Actually for me `commentstring` is set by vim-pandoc plugin so I'm just looking for a way to turn it off :-)

Anyway, you mean something like this:

```
if !exists('g:vimwiki_commentstring')
    setlocal commentstring=%%%s
elseif g:vimwiki_commentstring != ""                                                                                                                       
    setlocal commentstring=g:vimwiki_commentstring
endif
```
Unfortunately with `let g:vimwiki_commentstring='<!-- %s -->'` in my `.vimrc` I'm stuck with this error:

`E537: 'commentstring' must be empty or contain %s`

Do you have an idea?

tinmarino:
> Do you have an idea?

Yes (sorry for the delay)
 `setlocal commentstring`, like any set command,  should not get a variable like `g:vimwiki_commentstring` but a verbatim string like `<!-- %s -->` (without the quotes) 

So:
```
if !exists('g:vimwiki_commentstring')
    setlocal commentstring=%%%s
elseif g:vimwiki_commentstring != ""                                                                                                                       
    execute 'setlocal commentstring=' . g:vimwiki_commentstring
endif
```
will work

tinmarino:
Then please:
1. Add test
2. Rebase

tinmarino:
It's perfect !
Sorry for the travis fails: my bad: I change the test format, I'll reformat your test.

tinmarino:
I changed a little bit the code: Commit: 5651e744a64

1. We have a standardized way to declare variables (so they are all in the same place): have a look at vars.vim; `'commentstring': {'type': type(''), 'default': '%%%s'},`
2. We do not use `Include: vader_includes/vader_setup.vader`

Your feature is added in the dev branch, good job, see you in an other PR

Issue #951, opened 2020-07-24
Error when converting to html using custom script for markdown syntax


sdondley:
With `:Vimwiki2HTML` I get:
```
Error detected while processing function vimwiki#html#Wiki2HTML[1]..<SNR>112_convert_file:
line   27:
E121: Undefined variable: htmlfile
E15: Invalid expression: path_html.htmlfile
```

I fixed the issue with a hack. I have no idea if it's a correct hack, but hopefully it helps track down the problem. I'll submit pull request with the hack.

tinmarino:
Thanks for taking the time for the PR, it solved it (sorry for not merging your PR directly, I had some of mine to squash)
As I stole your commit, I offer you the command I discovered (may you already know):
```
git fetch wiki pull/953/head:test_953 # Steal the branch from your PR 
git checkout test_953  # checkout to the branch I decided to call test_953
```

Anyway, very nice catch.
Hope to see you soon.

PR #934, opened 2020-07-05
Generate RSS feed for diary

reinerh:
This change allows one to generate an RSS feed for the exported VimWiki (#172).
It introduces the command `:VimwikiRss` which generates a rss.xml file (filename can be changed via `rss_name` setting)
that contains the latest diary entries (by default 10; configurable via `rss_max_items`).

- [x] **ALL** pull requests should be made against the `dev` branch!
- [x] Take a look at [CONTRIBUTING.MD](https://github.com/vimwiki/vimwiki/blob/dev/CONTRIBUTING.md)
- [x] Reference any related issues.
- [x] Provide a description of the proposed changes.
- [x] PRs must pass Vint tests and add new Vader tests as applicable.
- [x] Make sure to update the documentation in `doc/vimwiki.txt` if applicable,
      including the Changelog and Contributors sections.

PR #953, opened 2020-07-24
prevent error when converting to html from markdown

sdondley:
See #951


tinmarino:
Thanks,
Error from:  #934
Fix: 21245a84358be7
which is what you gave but correcting for future potential brothers.

Issue #921, opened 2020-06-18
VimwikiGenerateTags and VimwikiDiaryGenerateLinks end up with empty wiki files


vrdel:
Prior to submitting a new issue make sure to complete these steps:

- [x] Checkout the `dev` branch and confirm the issue is present there as well.
      The `dev` branch contains fixes that may not have been merged to `master` yet.
- [x] Post the syntax you are using (default/mediawiki/markdown) **and** your vimwiki settings from your `.vimrc`
- [x] Provide a detailed description of the problem including **steps to reproduce the issue**.
- [x] Include the output of `:VimwikiShowVersion`.

Hi!
After update to new stable version v2.5, I cannot generate diary index nor list of tags with references anymore. Both functions delete all current entries in wiki file so that it ends up completely empty.

I'm using default syntax and relevant entries from my `.vimrc` wrt vimwiki configuration:
```
" vimwiki stuff
let g:vimwiki_folding='expr'
let wiki = {}
let wiki.path = '~/my_notes/wiki'
let wiki.path_html = '~/Dropbox/wiki_html'
let wiki.auto_export = 0
let wiki.template_path = '~/.vim/myscripts/'
let wiki.template_default = 'template'
let wiki.template_ext = '.tpl'
let wiki.nested_syntaxes = {'py': 'python', 'rb': 'ruby', 'pl': 'perl', 'c': 'c',
                    \ 'html': 'html', 'js': 'javascript', 'css': 'css', 'xml': 'xml',
                    \ 'sh': 'sh', 'zsh': 'zsh',
                    \ 'sed': 'sed', 'awk': 'awk', 'lex': 'lex',
                    \ 'markdown': 'markdown', 'otl': 'votl', 'tex': 'tex',
                    \ 'man': 'man', 'help': 'help', 'vim': 'vim', 'conf': 'config'}
let g:vimwiki_list = [wiki]
let g:vimwiki_ext2syntax = {'.wiki': 'media'}
let g:vimwiki_hl_headers = 1
let g:vimwiki_list_ignore_newline = 1
let g:vimwiki_table_auto_fmt = 0

function! VimwikiWikiIncludeHandler(value)
	pyfile ~/.vim/myscripts/includehandle.py
endfunction

function! VimwikiLinkHandler(link)
	pyfile ~/.vim/myscripts/linkhandle.py
endfunction

let g:vimwiki_toc_header = 'Sadrzaj'
```
[...]

PR #956, opened 2020-07-25
fix broken diary listing generation, #921

sdondley:
This seems to fix generation of diary index page.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1304,7 +1304,7 @@ function! vimwiki#base#update_listing_in_buffer(Generator, start_header,
         \ . s:safesubstitute(vimwiki#vars#get_syntaxlocal(header_level),
         \ '__Header__', a:start_header, '')
   keepjumps call append(start_lnum - 1, new_header)
-  let start_lnum += 1
+  let start_lnum += 2
   let lines_diff += 1
   if vimwiki#vars#get_wikilocal('syntax') ==# 'markdown'
     for _ in range(vimwiki#vars#get_global('markdown_header_style'))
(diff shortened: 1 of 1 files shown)


tinmarino:
Hi again @sdondley. I'm gonna code today. So will integrate (or not) this fast: let's work together!
Can you send me here some info for making a test ? File_pre : command : File_post that used to not work and now are working please: I am gonna do the test if you agree.

sdondley:
Sorry, I have no idea how write tests for vim and I barely know vimscript.

tinmarino:
After tests: if the chnages are only for Diary, do not modify update_buffer because all rely on its behavior: I dont think all are adding a counter bug before calling so the bug must be before in the call stack

tinmarino:
> Sorry, I have no idea how write tests for vim and I barely know vimscript.

That is why I am gonne code its: just past them here. I mean the :VimwikiDiaryGenerateLinks is working well for me so if you can reprodude, can you add some screenshots

sdondley:
diary generation

tinmarino:
@sdondley how do you add the keystrokes in the video ?

sdondley:
Screenflick. 

I did a little write up about looking for a good screen capture program: https://steve.dondley.com/Screen-capture-solutions/

tinmarino:
> Screenflick.

Thanks

You are at the good place in update_in_listing finally but it is a hard-to-read function.
The best would be to comment it a little.

Finally I did a hacky changing fold level at start and at end of this function. The commit reference the issue.

sdondley:
Thanks! It looks like it's working now. My listing no longer disappears when I do `<leader>wi`

tinmarino:
I felt free to close the PR

sdondley:
I am unable to close issue even on my own projects. I think there must be some kind of javascript issue with safari.

PR #963, opened 2020-07-31
Replace VimwikiListChangeLevel references in doc with VimwikiListChangeLvl

chipsenkbeil:
It's my first day trying out vimwiki, so bare with me if I'm missing something.

From what I can see, the command is defined as `VimwikiListChangeLvl` but is referenced in the documentation as `VimwikiListChangeLevel`. This updates the documentation. The other option would be to add some sort of alias to map VimwikiListChangeLevel -> VimwikiListChangeLvl, but doesn't seem like it'd be needed here.

Looks like my default formatting removed a couple of blank spaces. Not sure if I need to go back and address that or not.

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -840,7 +840,7 @@ Vimwiki file.
     Toggle checkbox of a list item disabled/off.
     See |vimwiki-todo-lists|.
 
-*:VimwikiListChangeLevel* CMD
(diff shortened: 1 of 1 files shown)


tinmarino:
> Looks like my default formatting removed a couple of blank spaces. Not sure if I need to go back and address that or not.

No, better like that.
Thank you and welcome

Issue #847, opened 2020-04-05
Support markdown multi-line highlight for constructs with inline-content


ubitux:
According to latest CommonMark spec as of `0.29` (https://spec.commonmark.org/0.29/#example-635), the following example should be highlighted:

```markdown
*hello
world*
```

With Vimwiki, the `conceal` is not effective here, which hints me that the highlight is not set (would be nice to have a different color for this markup by default).

Similarly, the following doesn't work (which is the one I use since it displays in a different color, even though the semantic is most of the time not correct):

```markdown
`hello
world`
```

Note that the spec says "Line breaks do not occur inside code spans", but this is only referring to the multi-line rendering: if you look at the output, it will still consider the whole thing as inline-code (ignoring the line break).

I use the `gq` command a lot, and thus the markup regularly breaks because of auto inserted line-breaks within highlighted sentences / word-groups.

**Note**: this issue is likely a good candidate for https://github.com/vimwiki/vimwiki/issues/664.

`:VimwikiShowVersion`:

```
Branch: dev
Revision: 78a41b7
Date: 2020-04-03 13:31:36 -0600
```

The issue is also reproducible with `master`.

tinmarino:
Feel free to make a PR. You can get it done in 2h.

chipsenkbeil:
@tinmarino, is this just syntax highlighting for these or is there extra logic to do?

If it's just syntax highlighting, then we could pop in a `syntax region ...` for each of these cases. Is the markdown supported by vimwiki following the commonmark spec? I don't actually know which flavor of markdown is used.

ubitux:
Speaking of this multi-line highlight not supported, there is a similar issue when using `let g:vimwiki_hl_cb_checked = 1`; if you have a multi line entry, only the first line gets highlighted when it's checked:

```markdown
- [x] hello
      world
```

(multiline is actually kind of supported, because when under "world", `o` will add a list entry, and `ctrl`-`space` works to toggle the entry as well)

I can open another issue if needed.

tinmarino:
@chipsenkbeil 
> @tinmarino, is this just syntax highlighting for these or is there extra logic to do?

According to the issue creator, it is just syntax highliting, for me too. But this may break the html conversion that is done line by line.
I never looked at html conversion, but a multiple line stuff is what you've just done that is why I called for your power.

@ubitux 
> I can open another issue if needed.

Yes it would be better: Small issues are easier to close. So it divides better the dev workflow. Please:

[...]

PR #967, opened 2020-08-01
Add inline comments

### LINKED ISSUE #965: [Feature Request] Inline & Multiline Comments
Another feature request here, not a bug. Making a note of this myself in case I'm able to dive in to offer this functionality. I'm also brand new to vimwiki, so there may be existing functionality that can do this or maybe people have existing workflows to achieve a similar result.

### Request

Support comments that are inline and/or multiline versus being at the start of a line.

```vimwiki
= My Wiki =

%% Line comment
- Item 1
- Item 2 %%+ Inline comment +%%
- Item 3
%%+ some comment
that continues
on multiple lines +%%
```

### Reason

I'm looking around at ways that other text-based languages handle extra metadata. Org mode has a variety of syntaxes like properties, tags, etc. that extensions like org-pomodoro leverage to mark tasks with metadata. The closest I've seen in vimwiki are tags, which also act as anchors; however, stashing extra metadata doesn't seem as approachable.

I'm not saying supporting metadata or any advanced functionality needs to be baked into vimwiki itself. Rather, providing inline comments means that plugins that want to extend vimwiki could stash extra info in the same line as a task, heading, etc. beyond the labelling that a tag provides. This also avoids tacking on information to the actual content when it's rendered into HTML or other formats where you don't want it to be included.

### Example Use Case

If I was to implement a plugin that extends vimwiki to support the pomodoro technique, one way would be to do the following:

1. Detect if cursor is on a task
2. Prepend some tags to indicate being used with pomodoro for searching/identification
3. Append an inline comment that contents metadata associated with the task, such as the start time, time spent, or pomodoro achieved

```vimwiki
- [ ] :pomodoro:active: Some task that is actively being tracked %%+ {"time": "10 minutes"} +%%
- [ ] :pomodoro: Some task that has been tracked, but is not active and not completed %%+ {"time": "7 minutes"} +%%
- [X] :pomodoro:complete: Some task that has been completed %%+ {"time": "38 minutes", "pomodoro_count": "1" } +%% 
- [ ] Some task not being tracked for time
```

Of course, the plugin would collapse the metadata to hide it (similar to how links are collapsed until you move over the line.

```vimwiki
- [ ] :pomodoro:active: Some task that is actively being tracked %%+ ... +%%
- [ ] :pomodoro: Some task that has been tracked, but is not active and not completed %%+ ... +%%
- [X] :pomodoro:complete: Some task that has been completed %%+ ... +%% 
- [ ] Some task not being tracked for time
```

This can be done today with pure tags (if you don't need specific details), line comments above/below and some fancy wizardry, or using a separate database in combination with tags to keep track of which task matches which subset of data. Seems like being able to try out new features by encoding information in inline comments could open up some interesting doors.

### Requirements

Minimum for my use case would be the ability to add line comments later in a line.

```vimwiki
- Some text %% comment rest of line
```

Full-blown solution would be able to use comments with a start and stop breakdown.

```vimwiki
- Some text %%+ inline comment +%% more text
```

chipsenkbeil:
Resolves #965.
### Details

Support comments that are inline and/or multiline versus being at the start of a line.

```vimwiki
= My Wiki =

%% Line comment
- Item 1
- Item 2 %%+ Inline comment +%%
- Item 3
%%+ some comment
that continues
on multiple lines +%%
```

### Status

- [X] Add `rxMultilineComment` syntax for default vimwiki (markdown and mediawiki are empty)
    - [X] Syntax test for comments (single-line & multiline) added to `vimwiki/test/syntax.vader` and passing
- [X] Add multiline comment parsing in `s:parse_line` of `vimwiki/autoload/vimwiki/html.vim`
    - [X] Comment parsing test for comments (single-line & multiline) added to `vimwiki/test/convert_default_html.vader` and passing
- [X] Update changelog to reflect new PR #967

### DIFF
--- a/syntax/vimwiki_default.vim
+++ b/syntax/vimwiki_default.vim
@@ -95,6 +95,8 @@ let s:default_syntax.rxPreEnd = '}}}'
 let s:default_syntax.rxMathStart = '{{\$'
 let s:default_syntax.rxMathEnd = '}}\$'
 
+let s:default_syntax.rxMultilineCommentStart = '%%+'
+let s:default_syntax.rxMultilineCommentEnd = '+%%'
 let s:default_syntax.rxComment = '^\s*%%.*$'
 let s:default_syntax.rxTags = '\%(^\|\s\)\@<=:\%([^:''[:space:]]\+:\)\+\%(\s\|$\)\@='
 
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1245,6 +1245,7 @@ function! s:parse_line(line, state) abort
   let state.para = a:state.para
   let state.quote = a:state.quote
   let state.arrow_quote = a:state.arrow_quote
+  let state.active_multiline_comment = a:state.active_multiline_comment
   let state.pre = a:state.pre[:]
   let state.math = a:state.math[:]
   let state.table = a:state.table[:]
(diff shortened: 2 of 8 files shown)


tinmarino:
Pretty good ! Thanks for the tests
Announce when your dev is finished so that we can press the merge button.

syntax/vimwiki_default.vim
```diff
@@ -95,7 +95,12 @@ let s:default_syntax.rxPreEnd = '}}}'
 let s:default_syntax.rxMathStart = '{{\$'
 let s:default_syntax.rxMathEnd = '}}\$'
 
-let s:default_syntax.rxComment = '^\s*%%.*$'
```
tinmarino, line 98: I think this is breaking the tests.
It is nice that you can add your feature without adding a new confguration variable but extending the existing one: there are too many variables ...

chipsenkbeil: Yeah, still working on the tests. :) Was building the docker images so I can run locally first.

Unless there are tests that expect `%%+` to be a single-line comment, this _shouldn't_ break things, but it is vimscript, so you never know! That's the only part that isn't backwards compatible is that anyone who has a comment that is `%%+ANYTHING ELSE` and assuming it will be a single-line comment will suddenly have it fail.

My thoughts are that we introduce this as a breaking change to everyone as I can't see `%%+` being a common start to a comment today. We could have `g:vimwiki_enable_multiline_comments` as a variable that could be set to 0 to disable this feature for backwards compatibility.

Personally, I'm a fan of rolling it out to everyone. :)

chipsenkbeil: Actually, I may not even need that change to rxComment based on how I ended up implementing this. Let me try reverting that part. :)

chipsenkbeil:
@tinmarino, ready for review and merge when ready.

tinmarino:
Wonderfull ! 
Very nice contribution, thank you.

If you have time, maybe you can get a glimpse at multiline highlighting there: #847
For your the API with your plugin, I'd be glad to help, so feel free to create a separete issue in time.

See you.

tinmarino:
+1 for the explicative comments
+1 for the self made tests (its hard but for me too :)

Issue #928, opened 2020-06-26
Changing the delimiter of the tag


Tecvoc:
I have been using a custom script along with pandoc for converting vimwiki in markdown format to html, along with emojis. However emojis have a syntax very similar to tags ( like :star : for :star:) and they create new tags.

Is there is a way we can set the delimiter in the vimwiki global options for tags.

Currently I have identified as to where the change are needed, in order to do so, however due to limited vimscript knowledge I am just posting the diff code that worked for me.

```vimscript
syntax/vimwiki_markdown.vim

-let s:markdown_syntax.tag_search = '\(^\|\s\)\zs:\([^:''[:space:]]\+:\)\+\ze\(\s\|$\)'
+let s:markdown_syntax.tag_search = '\(^\|\s\)\zs@\([^@''[:space:]]\+@\)\+\ze\(\s\|$\)'

autoload/vimwiki/tags.vim
-     for tag in split(tag_group, ':')
+     for tag in split(tag_group, '@')

```

The above changes the delimiter to '@' like @mytag1@.

tinmarino:
Related:  #922

tinmarino:
Fixed: 188ead55db340f

See the new doc: https://github.com/vimwiki/vimwiki/blob/188ead55db340fc10e9d5facf40f0ecb42b81985/doc/vimwiki.txt#L3602-L3631

In your case ( after rebase on dev ):

```vim
let g:vimwiki_tag_format = {'pre_mark': '@', 'post_mark': '@', 'sep': '@'}
```

PR #976, opened 2020-08-07
Feature: Add support for emoji conceal and complete

tinmarino:
Emoji are concealed : `:cat:` -> 🐱 and completed `i<C-X><C-U>` all the code was hard copied from https://github.com/junegunn/vim-emoji
Doc contains an emoji map for convenience (I always use my vimwiki's one, think it may be usefull to other)
Here after some relevant links:

* Data:
  * Doc: https://github.com/onmyway133/emoji/blob/master/README.md
  * CheatSheet (nice with browser): https://www.webfx.com/tools/emoji-cheat-sheet/
  * Visual: https://en.wikipedia.org/wiki/List_of_Unicode_characters#Number_Forms
  * Full Txt: https://ftp.unicode.org/Public/emoji/13.0/emoji-sequences.txt
  * Md inteligent "fake link", easy to use: https://github.com/swharden/code-notes/blob/master/Csharp/emoji.md

* See also (maybe to put in doc to advice user)
  * https://github.com/chrisbra/unicode.vim
  * https://github.com/ryanoasis/vim-devicons
  * https://github.com/lervag/vimtex


tinmarino:
The next task after that is to make an option for tag delimiter  #928

PR #1008, opened 2020-09-05
Fix: HTML Conversion: List with wrapped lines

### LINKED ISSUE #1007: Regression in HTML Blockquote conversion
I think a recent change to HTML blockquote conversion on `dev` (https://github.com/vimwiki/vimwiki/commit/dc63a5dacc97d957f35995b1944a3a6a26647d85) may have caused a regression when converting a file that contains a list element that hard wraps to a second line. 

For example, say I have a file with:

```
  - bullet with long text that keeps going and eventually wraps around should
    not create a blockquote
```

Here's the rendered HTML pre/post the mentioned commit:
DeepinScreenshot_select-area_20200903093053

I haven't yet looked into the code to see what a potential fix might look like, first I wanted to check to see if this change is intentional or in fact a regression.

Thanks!

-----

Os: Linux
Vim: 801
Branch: dev
Revision: dbbadc8
Date: 2020-09-01 20:13:12 -0400

-----

Prior to submitting a new issue make sure to complete these steps:

- [x] Checkout the `dev` branch and confirm the issue is present there as well.
      The `dev` branch contains fixes that may not have been merged to `master` yet.
- [x] Post the syntax you are using (default/mediawiki/markdown) **and** your vimwiki settings from your `.vimrc`
- [x] Provide a detailed description of the problem including **steps to reproduce the issue**.
- [x] Include the output of `:VimwikiShowVersion`.

rylwin:
Track how many leading spaces a list item has in order to determine
whether we should start a blockquote.

Note the extra close tag in the added test: this is a pre-existing issue
already present on `dev`.

Resolve #1007 (or at least part of it)

-----

----

Not sure if I should update changelog, since this is a fix for a change that hasn't made it to master yet -- let me know.

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1005,7 +1005,7 @@ function! s:process_tag_arrow_quote(line, arrow_quote) abort
 endfunction
 
 
-function! s:process_tag_list(line, lists) abort
+function! s:process_tag_list(line, lists, lstLeadingSpaces) abort
   function! s:add_checkbox(line, rx_list) abort
     let st_tag = '<li>'
     let chk = matchlist(a:line, a:rx_list)
@@ -1026,6 +1026,7 @@ function! s:process_tag_list(line, lists) abort
 
 
   let in_list = (len(a:lists) > 0)
+  let lstLeadingSpaces = a:lstLeadingSpaces
 
   " If it is not list yet then do not process line that starts from *bold*
   " text.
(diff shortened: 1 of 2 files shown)


autoload/vimwiki/html.vim
```diff
@@ -1323,6 +1332,7 @@ function! s:parse_line(line, state) abort
   let state.quote = a:state.quote
   let state.arrow_quote = a:state.arrow_quote
   let state.active_multiline_comment = a:state.active_multiline_comment
+  let state.list_leading_spaces = a:state.list_leading_spaces
```
rylwin, line 1335: I wanted to avoid adding to the state, but couldn't think of a better way to determine whether we want to be in a blockquote than to keep track of the leading spaces for the current list item. Open to other ideas!

And now that I'm reflecting more, I wonder if leading spaces is a bit misleading, as we also include the bullet/number and checkbox if there is one, so perhaps a name like "list_text_start_column" might be a better description?

tinmarino: `let state.list_leading_spaces` is a perfect variable.
About adding new state, it is ok, it may be the best way to do that

tinmarino (approved):

tinmarino:
Perfect! Thank you for the PR (with tests!)

Feel free to open new ones: I try to merge them as soon as possibe to avoid conflicts

Issue #1021, opened 2020-09-23
UltiSnips <tab> doesn't work while using vimwiki (related issues: #357 #669)


MahbubAlam231:
My vimwiki syntax:
```vim
let g:vimwiki_ext2syntax = {'.md': 'markdown'}
let g:vimwiki_list = [{'path': '~/vimwiki/',
            \'syntax': 'markdown',
            \'ext': '.md'}]
```
`:VimwikiShowVersion`: `Stable version: 2.5`

Related issues: #357 #669 

**Description of the problem:**
[UltiSnips](https://github.com/SirVer/ultisnips) `<tab>` doesn't work with vimwiki filetypes. `:verbose imap <tab>` shows:
```vim
i  <Tab>       *@vimwiki#tbl#kbd_tab()
        Last set from ~/.vim/plugged/vimwiki/ftplugin/vimwiki.vim line 543
i  <Tab>       * <C-R>=UltiSnips#ExpandSnippet()<CR>
        Last set from ~/.vim/plugged/ultisnips/autoload/UltiSnips/map_keys.vim line 61
```
On the other hand there are plugins such as [supertab](https://github.com/ervandew/supertab) whose `<tab>` works along side UltiSnips. I wanted to keep the behaviour of `<tab>` from UltiSnips, so to overcome this issue I put the following in my `.vimrc`:
```vim
let g:vimwiki_table_mappings=0
augroup ChangeVimwikiTableTab
    autocmd!
    autocmd Filetype vimwiki inoremap <silent><expr><buffer> <M-n> vimwiki#tbl#kbd_tab()
    autocmd Filetype vimwiki inoremap <silent><expr><buffer> <M-p> vimwiki#tbl#kbd_shift_tab()
augroup end
```

According the `:h vimwiki` [here](https://github.com/vimwiki/vimwiki/blob/619f04f89861c58e5a6415a4f83847752928252d/doc/vimwiki.txt#L3617), `g:vimwiki_table_mappings` option will be fully removed at a later point. Then the above wouldn't work.

I think UltiSnips is a very useful plugin. I wish vimwiki `<tab>` worked alongside UltiSnips `<tab>` just like [supertab](https://github.com/ervandew/supertab) does -- is there any way to implement this from `.vimrc` or elsewhere?

----------------------
Prior to submitting a new issue make sure to complete these steps:

- [X] Checkout the `dev` branch and confirm the issue is present there as well.
[...]

PR #1030, opened 2020-10-09
Allow overwriting insert mode maps

davidsierradz:
According to `:h hasmapto()` if we don't pass the mode as an argument then only normal, visual and operator-pending modes are checked.

Lets pass the mode explicitly to allow overwriting insert mode mappings.

I think this will fix: #1021 

Original PR that introduced `vimwiki#u#map_key()`: #686 

With this minimal `.vimrc | init.vim`:

```viml
call plug#begin('~/.config/nvim/plugged')
Plug 'vimwiki/vimwiki', { 'branch': 'dev' }
call plug#end()
let g:vimwiki_list = [
      \ {'path': '~/notes/',
      \ 'syntax': 'markdown', 'ext': '.md'}]
imap <F13> <Plug>VimwikiIncreaseLvlSingleItem
inoremap <C-T> <Esc>
```

Notice I want to remap `<C-T>` in insert mode. We should only need to remap `<Plug>VimwikiIncreaseLvlSingleItem` to another key (`<F13>` in this case) and vimwiki should not apply this: `call vimwiki#u#map_key('i', '<C-T>', '<Plug>VimwikiIncreaseLvlSingleItem')` from `ftplugin/vimwiki.vim`.

Without this patch this is not happening, `:verbose imap <C-T>`:

image

With the patch:

image

I searched all vimwiki codebase and noticed all `vimwiki#u#map_key` calls always have the mode as first argument, so I think this is a safe fix. Writing a test is above my knowledge/time.

### DIFF
--- a/autoload/vimwiki/u.vim
+++ b/autoload/vimwiki/u.vim
@@ -212,7 +212,7 @@ endfunc
 function! vimwiki#u#map_key(mode, key, plug, ...) abort
   if a:0 && a:1 == 2
     " global mappings
-    if !hasmapto(a:plug) && maparg(a:key, a:mode) ==# ''
+    if !hasmapto(a:plug, a:mode) && maparg(a:key, a:mode) ==# ''
       exe a:mode . 'map ' . a:key . ' ' . a:plug
     endif
   elseif a:0 && a:1 == 1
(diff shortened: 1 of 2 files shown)


tinmarino:
Wonderfull!! => this will actually remove many bugs <= no plugin should overwrite the other without user consentment: ticket statistic is demonstrating that.

This does not "fix" #1021 because #1021 is a desired behavior. See my comment there.

Thank you!

tinmarino:
Do you want me to add your name in the contributor list?
```
  - David Sierra DiazGranados (@davidsierradz)
```

PR #1039, opened 2020-10-22
fix: not throwing error when loading for some languages

### LINKED ISSUE #1029: E806: using Float as a String
When I do a "git commit" using vim as editor I get a vimwiki error

Hinweis: Warte auf das Schließen der Datei durch Ihren Editor... Error detected while processing /Users/peter/.vim/plugged/vimwiki/plugin/vimwiki.vim:
line   13:
E806: using Float as a String

As a workaround I set the version number to dev as described in the source code on line 13:

" Set to version number for release, otherwise -1 for dev-branch
let s:plugin_vers = -1

Is there a better fix?

dmouraneto:
Fixes #1029 
Some languages uses comma instead of dot for decimal separator.
When loading vimwiki the following error was occuring
E806: using Float as a String
This is probably a bug in vim, as a workaround when using float can be
str2float.
Similar issue in another project
https://github.com/nathanaelkane/vim-indent-guides/issues/10

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -3830,6 +3830,7 @@ Contributors and their Github usernames in roughly chronological order:
     - Jerome Gay (@jeromg)
     - Benney Au (@chinwobble)
     - David Sierra DiazGranados (@davidsierradz)
+    - Daniel Moura (@dmouraneto)
 
 ==============================================================================
 16. Changelog                                              *vimwiki-changelog*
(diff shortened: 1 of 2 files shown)


tinmarino:
Hi @dmouraneto 

Thank you for the contribution.

I thought maybe we could change the local in vim but this would break many many other things, so this PR fixes at least one issue.

Good job.

PR #1047, opened 2020-11-08
Allow VimwikiToggleListItem mapping to be replaced

deepredsky:
Currently VimwikiToggleListItem mapping cannot be removed but only new mapping
to it can be added. This commit will allow to add custom binding to this
function and release default mapping of `<C-space>` and `<C-@>`, which was how it
was before https://github.com/vimwiki/vimwiki/pull/686

Steps for submitting a pull request:

- [X] **ALL** pull requests should be made against the `dev` branch!
- [x] Take a look at [CONTRIBUTING.MD](https://github.com/vimwiki/vimwiki/blob/dev/CONTRIBUTING.md)
- [X] Reference any related issues.
- [X] Provide a description of the proposed changes.
- [x] PRs must pass Vint tests and add new Vader tests as applicable.
- [X] Make sure to update the documentation in `doc/vimwiki.txt` if applicable,
      including the Changelog and Contributors sections.

PR #1048, opened 2020-11-09
enable changing insert mode table mappings

amitbeka:
Mapping `<Tab>` and `<S-Tab>` at insert mode sometimes collide with
other plugins which utilize these for completions (e.g. coc example
mapping).

Since for many users the case for completion is more frequent than table
editing, this patch enables changing these mappings using the options
`g:vimwiki_tbl_next_cell` and `g:vimwiki_tbl_prev_cell`.

This patch makes it easier of those using multiple plugins and have no
desire or knowledge in debugging the collision.
It complements #1030 by providing a more direct and documented method
for setting custom keymappings.

Signed-off-by: Amit Beka <--->

- [X] PRs must pass Vint tests and add new Vader tests as applicable.

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -612,7 +612,14 @@ TABLE MAPPINGS, INSERT MODE                           *vimwiki-table-mappings*
 
                                                        *vimwiki_i_<Tab>_table*
 <Tab>                   Go to the next table cell, create a new row if on the
-                        last cell.
+                        last cell. Adjusts the whole column width according
+                        to the largest cell.
+                        Remap command: `<Plug>VimwikiTableNextCell`
+
+                                                       *vimwiki_i_<Tab>_table*
+<S-Tab>                 Go to the previous table cell. Adjusts the whole
+                        column width according to the largest cell.
+                        Remap command: `<Plug>VimwikiTablePrevCell`
 
 LIST MAPPINGS, INSERT MODE                             *vimwiki-list-mappings*
                                                               *vimwiki_i_<CR>*
@@ -3844,6 +3851,7 @@ Contributors and their Github usernames in roughly chronological order:
     - Daniel Moura (@dmouraneto)
     - Tomáš Janoušek (@liskin)
     - Rajesh Sharem (@deepredsky)
+    - Amit Beka (@amitbeka)
 
 ==============================================================================
 16. Changelog                                              *vimwiki-changelog*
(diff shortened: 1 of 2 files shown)


tinmarino:
Hello,

Thank you for the PR. Some changes must be made because this is a PR for "those who want to remap the <Tab> key for table navigation" and we cannot treat all those who want to remap keys specifically (even if in this case they are already lol). You can improve the table tab key magement alongside ALL the others which would make your PR much more usefull and easy to maintain

1/ Look at the way to add configuration variable in vars.vim (it is centralized => Much easier to test and maintain) (we cannot accept a PR that have a config not in this file.
2/ Look at vimwiki#u#map_key other key mapping are using
3/ Make table use it
4/ Make it configurable like remap <tab> -> >>

Otherwise, you can keep all this configuration in your vimrc and paste here the code for others to copy paste it

tinmarino (changes requested):
See comment

amitbeka:
Sure, I can change these. I had an impression that `vimwiki#u#map_key` won't handle this, because it is currently used only for `<Plug>` commands, so I was unsure if it was by design.

What exactly do you mean by "make it configurable like remap -> >>"? Not sure I've understood.

tinmarino:
> What exactly do you mean by "make it configurable like remap -> >>"? Not
sure I've understood.

To have something like a dictionarry so that you can change mappings to
remap tab to >> so that >> is doing what tab is suposed to do.
Or to choose directly the keystoke to map to a certain function.

tinmarino:
In the function, you read the user dic and if the left side (keystroke) or
rhe right side (fubction) is in the dic, tou use the value from the dic
instead of the default.
After this is done, default can be set in this dic in vars.vim for example.
And user can change them before loading vimwiki

tinmarino:
You can have a look at #1047  and `vimwiki#u#map_key`

amitbeka:
Thanks, I hope to get back to this PR today or tmrw

amitbeka:
Done fixing, would be happy for more comments

tinmarino:
@amitbeka sorry I did not notice your comment (until you request review which was a good idea).

This PR improves the code because:
1/ It is more table, easier to refactor
2/ Doc
3/ Very easy to disable the map (2 ways)
4/ Very easy to remap: Read the doc.

I would have been glad to see some test, but I'll keep a ref to it and will do it with some others.
Anyway, thank you.

physicophilic:
Hi.. how can I use this feature? I have the latest release, but this is a newer feature (and I would love it).

tinmarino:
@physicophilic you have to "checkout to the dev branch"
Go to the folder where your vimwiki plugin is installed and run something like `git checkout origin/dev` or ask you plugin manager to follow the dev branch and not the master one.

physicophilic:
Thank you. Setting the dev branch from the plugin manager worked. I couldn't make git commands work though; not familiar with git.

PR #1051, opened 2020-11-12
Fix Renaming issue in Windows and ext with two dot

levirs565:

In this Pull Request i fix two bugs:
- Rename not work in Windows. Fixed with replace `\` in path with `/`
- Updating link not work when wiki extension contain two dot (eg. `.wiki.md`). Fixed with using regex rather than using `fnamemodify`/`expand`.

I also refractor `vimwiki#base#rename_link` in order to simply the code and make it less bugs.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -62,12 +62,18 @@ function! vimwiki#base#subdir(path, filename) abort
     let filename = a:filename
   endif
   let idx = 0
+  let pathelement = split(path, '[/\\]') 
+  let fileelement = split(filename, '[/\\]')
+  let minlen = min([len(pathelement), len(fileelement)])
   "FIXME this can terminate in the middle of a path component!
-  while path[idx] ==? filename[idx]
+  while pathelement[idx] ==? fileelement[idx]
     let idx = idx + 1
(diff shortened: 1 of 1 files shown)


tinmarino:
@levirs565 
Thanks for the PR, it seems nice.
May you comment, give a non-regression test?

This forces the '/' delimiter even on Windows. I am in favor of it anyway: it is more transportable

levirs565:
What do you mean?. I am have bad english grammar.

levirs565:
How do i run test?

tinmarino:
You write here an example of a problem you solve.
For example with this file (copy rhe file) I press enter and It does not go to the good file but Vim gives the error (copy the error)

levirs565:
Problem that fixed with this pull request:
1. Renaming page failed in Windows
2. Updating link (after renaming page) failed when wiki extension contain two dot (eg `.wiki.md`)

tinmarino:
Thank you

levirs565:
This pull request now only fix bug renaming in Windows. I have delete the commit that fix bug updating link in two dot ext. I also delete commit that update `doc/vimwiki.txt`. **This pull request is not finished**

PR #1056, opened 2020-11-20
Revert "Fix Renaming issue in Windows"

levirs565:
Reverts vimwiki/vimwiki#1051. My Pull Request is not finished. I have deleted commit that change `doc/vimwiki.txt`. It means the `doc/vimwiki.txt` is **not updatted**. And Also the commit name is wrong. The pull request **only fix Reaming Page in WIndows not fix renaming in two dot extension**. Two dot extension file is **bad** and it is **not standard**.


tinmarino:
Sorry, my bad. If you cannot reopen  #1051, just open an other PR.

PR #1073, opened 2020-12-29
Add option to configure date string format

yhu266:
This PR adds option `template_date_format` to configure string format of `%date%` in html templates.

---

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -3889,8 +3889,6 @@ http://code.google.com/p/vimwiki/issues/list. They may be accessible from
 https://github.com/vimwiki-backup/vimwiki/issues.
 
 New:~
-    * Feature: Add option |vimwiki-option-template_date_format| to configure
-      alternative date string format
     * Feature: PR #686: add a flag lists_return to disable mappings 
       to <CR> and <S-CR> in insert mode
     * Feature: enable re-mapping insert mode table mappings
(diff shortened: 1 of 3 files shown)


doc/vimwiki.txt
```diff
@@ -3912,6 +3925,8 @@ New:~
     * PR #907: Cycle bullets
     * PR #901: adds multiparagraph blockquotes using email style syntax
     * PR #934: RSS feed generation for diary with :VimwikiRss.
+    * Feature: Add option |vimwiki-option-template_date_format| to configure
```
tinmarino, line 3928: Just a note: I try to put new changes above so that history goes from bottom to top. But usually contributors, like you,  put it below the paragraph, so it does not matter.

tinmarino:
Hi @yhu266 it seems good. This is a nice option for a test. I will do the Vader code but can you paste here an example of use case (before, command, after) to help the test coding please?

yhu266:
The function `process_date` uses `strftime` to convert placeholder `%date%` to a string of current date, if a fixed string is not specified in `wiki` files. Thus any valid input for `strftime` is valid for the option `template_date_format`.

Short examples:
`%b %d, %Y` will produce `Dec 29, 2020`,
`%c` will produce `Tue Dec 29 05:52:11 2020`,
`%y%m%d %T` will produce `201229 05:53:41`,
and so on.


Full example:

**configurations**
```vim
let wiki = {}
let wiki.template_date_format = '%b %d, %Y'
let g:vimwiki_list = [ wiki ]
```

**template**
```html
<html>
<body>
    <div class="content">
        <p><small>Last updated on %date%</small></p>
    </div>
</body>
</html>
```

and the generated `html` for today (December 29, 2020) will be
```html
<html>
<body>
    <div class="content">
        <p><small>Last updated on Dec 29, 2020</small></p>
    </div>
</body>
</html>
```

tinmarino:
Prefect!

You can now enjoy your new feature in the dev branch.

Thank you @yhu266.

PR #884, opened 2020-05-18
Add a diary frequency option

jeremcs:
This PR is a draft to implement different diary frequencies.
This feature has been in the issue tracker for [quite some time](https://github.com/vimwiki/vimwiki/issues/265).

Basically, it provides two new fields for wikis : `diary_frequency` and `diary_start_week_monday`. They respectively default to `daily` and `1` (true).
```vim
let g:vimwiki_list = [
    \ {'path': '~/vimwiki/weekly', 
    \  'diary_frequency': 'weekly',
    \  'diary_start_week_monday': 1}
    \]
```

I initially had more ambitions for this PR, but I'll stick to adding a weekly period for now.

As the end user, when you open today's entry, you open this week's entry.
It is represented by the first day of the week, so for this week, the file is named `2020-05-18.<your_ext>`.
Whether it's a Monday or a Sunday is decided by the `diary_start_week_monday` flag.

Likewise, yesterday's diary is last week's diary and tomorrow is next week's.

As you can see, this is basically done by changing the `vimwiki#diary#diary_date_link` to accept a reference timestamp and how many days or weeks (according to this diary's settings) we shift.

What do you think ?

Also, thanks for the great plugin!

-----

Steps for submitting a pull request:

- [x] **ALL** pull requests should be made against the `dev` branch!
- [x] Take a look at [CONTRIBUTING.MD](https://github.com/vimwiki/vimwiki/blob/dev/CONTRIBUTING.md)
- [x] Reference any related issues.
- [x] Provide a description of the proposed changes.
- [x] PRs must pass Vint tests and add new Vader tests as applicable.
  No errors _added_. (I already have errors on dev when running locally. I don't know if it's due to my setup.)
- [x] Make sure to update the documentation in `doc/vimwiki.txt` if applicable, including the Changelog and Contributors sections.

PR #1082, opened 2021-01-12
Default to wiki number 0 if diary is called outside a wiki buffer

neg:
If a diary page is opened from outside a wiki buffer the diary frequency
and start of week options are not respected as the wiki number is not
yet set for the local buffer.

Fix this by defaulting the wiki number to 0, which is what is done in
vimwiki#diary#make_note() for the diary_rel_path option that suffers
from the same problem.

Example on how to trigger the problem:

    let g:vimwiki_list = [{'path': '~/.vimwiki', 'diary_frequency': 'weekly', 'diary_start_week_day': 'monday'}]

    $ vim -c VimwikiMakeDiaryNote

Fixes: a3be479d5a5fc290 ("Add a diary frequency option (#884)")
Signed-off-by: Niklas Söderlund <niklas.soderlund@ragnatech.se>

### DIFF
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -37,6 +37,17 @@ endfunction
 
 function! vimwiki#diary#diary_date_link(...) abort
   " Return: <String> date
+
+  let wiki_nr = vimwiki#vars#get_bufferlocal('wiki_nr')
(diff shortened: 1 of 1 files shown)


tinmarino (approved):
Seems good, as it opens the first wiki, better having its default conf

tinmarino:
Thank you @neg

Issue #1061, opened 2020-12-04
<C-Space> no longer toggles todo list items


brennen:
Latest change on dev branch. First noticed in default syntax, but also broken for Markdown.

Create a TODO list item like so:

```vimwiki
  - [ ] foo
```

**Expected behavior:** Hitting Ctrl-Space with the cursor on that line should toggle the state of the checkbox.

**What happens:** It doesn't toggle. I haven't had time to dig any further, so I don't know if it's not firing `:VimwikiToggleListItem` or if `:VimwikiToggleListItem` is itself broken.

**:VimwikiShowVersion output**

```
Os: Linux
Vim: 801
Branch: dev
Revision: 8d4cb7f
Date: 2020-11-20 11:05:51 -0300
```

**vimrc**

```vim
" a stub vimrc for just running vimwiki on an otherwise stock setup

set nocompatible
set nomodeline
let mapleader = ","

" vundle setup & vundle-managed plugins {{{

  filetype off

  set rtp+=~/.vim/bundle/Vundle.vim/
  call vundle#begin()

  " let Vundle manage Vundle - required:
  Plugin 'VundleVim/Vundle.vim'

  " CUSTOMIZE: vimwikis - for notes, daily logs, etc. {{{

    " main personal wiki
    let wiki = {}
    let wiki.path = '~/notes/vimwiki/'
    let wiki.path_html = '~/notes/html/'
    let wiki.auto_tags = 1
    let wiki.auto_diary_index = 1
    let wiki.automatic_nested_syntaxes = 1

    " a markdown wiki for testing purposes
    let markdownwiki = {}
    let markdownwiki.path = '~/notes/markdownwiki'
    let markdownwiki.path_html = '~/notes/markdownwiki/html/'
    let markdownwiki.auto_tags = 1
    let markdownwiki.auto_diary_index = 1
    let markdownwiki.automatic_nested_syntaxes = 1
    let markdownwiki.syntax = 'markdown'
    let markdownwiki.ext = '.md'

    " there can be many of these
    let g:vimwiki_list = [wiki, markdownwiki]

    " don't make temporary wikis based on file extensions in the list - this
    " is necessary to avoid .md files getting a filetype of vimwiki instead of
    " markdown:
    let g:vimwiki_global_ext = 0

    let g:vimwiki_folding = 'expr'
    " let g:vimwiki_folding = 'syntax'
    " let g:vimwiki_folding = 'list'

    " links concealing, etc.
[...]

PR #1083, opened 2021-01-13
Restore <C-@> list item toggle on Unix systems

neg:
A changed aimed to allow user configuration files to be able to override
all default mappings for the VimwikiToggleListItem plug did so by
removing the optional forth argument to vimwiki#u#map_key(). This
argument if set to 1 allows the same plug to be mapped to multiple keys
in the same mode. If not present only the first mapping takes effect and
all later attempts to map the plug in the same mode result in a no-op.
Thus allowing any user configuration to set it first and then 'skipping'
the defaults.

The change however did not account for that the default on Unix systems
was to map both <C-Space> _and_ <C-@> to VimwikiToggleListItem so the
change breaks this behavior, described in issue #1061.

Fix this by restoring the forth argument to vimwiki#u#map_key() to allow
multiple mappings to the same plug/mode but wrap it in a block to first
check if the user have overridden it or not. This goes back to how his
was handled before vimwiki#u#map_key() was added in [1].

1. 4106cb7bc739a45f ("New option g:vimwiki_key_mappings to enable/disable key mappings.")

Fixes: 48baa1f4cd1bb496 ("Allow VimwikiToggleListItem mapping to be replaced (#1047)")
Signed-off-by: Niklas Söderlund <niklas.soderlund@ragnatech.se>

### DIFF
--- a/ftplugin/vimwiki.vim
+++ b/ftplugin/vimwiki.vim
@@ -509,11 +509,13 @@ nnoremap <silent><buffer> <Plug>VimwikiListO
 " Declare Map: default lists key mappings (again)
 if str2nr(vimwiki#vars#get_global('key_mappings').lists)
   call vimwiki#u#map_key('n', 'gnt', '<Plug>VimwikiNextTask')
-  call vimwiki#u#map_key('n', '<C-Space>', '<Plug>VimwikiToggleListItem')
-  call vimwiki#u#map_key('v', '<C-Space>', '<Plug>VimwikiToggleListItem')
-  if has('unix')
-    call vimwiki#u#map_key('n', '<C-@>', '<Plug>VimwikiToggleListItem')
-    call vimwiki#u#map_key('v', '<C-@>', '<Plug>VimwikiToggleListItem')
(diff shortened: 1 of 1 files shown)


neg:
I'm a bit confused if anything is expected of me to clear the last checkbox. This is a simple fix for a regression not yet released so I see no need to touch the documentation or changelog. For the Contributors section, sure I could add myself but as I'm only drive-by fixing an issue that hit my workflow I would be perfectly happy without doing so, but if it's needed (or something else is) to get this merged please let me know.

tinmarino:
Hi @neg 

> I'm a bit confused if anything is expected of me to clear the last checkbox

Nop, I made a fast review (clicking the buttons) then, left time for other people to review
It does not bring any problems so I'll merge.

Thank for your contribution.

tinmarino:
Let's add your name as contributor with something you will be proud of.

You can have a look at `good first issue` label.

brennen:
Confirmed this does indeed fix the regression described at #1061. Thanks much for the fix!

PR #1110, opened 2021-03-29
Vimwiki2HTML: remove string concat from variable substitution

brennen:
Problem: I noticed that running :Vimwiki2HTML for a file with a `"` in the `%title` line would error out.  For example, a title like:

```vimwiki
%title this will die: "
```

Causes:

```
Error detected while processing function vimwiki#html#Wiki2HTML[1]..<SNR>177_convert_file[15]..<SNR>177_convert_file_to_lines_template:
line    8:
E116: Invalid arguments for function substitute(v:val, "%title%", "this will die: "", "g")
```

Solution: It seems like the string concatenation here was unnecessary. At least on my vim 8.1, the evaluated code seems to have access to the correct variables.

I'm sure there are bigger problems with the HTML generation here, but this allows my ~2000 pages with titles to render without throwing errors.

Also adds a brief test and removes some unnecessary DeleteFile() calls from html_convert_default.vader.

- [X] Reference any related issues. (Didn't find anything.)

### DIFF
--- a/autoload/vimwiki/html.vim
+++ b/autoload/vimwiki/html.vim
@@ -1831,25 +1831,25 @@ function! s:convert_file_to_lines_template(wikifile, current_html_file) abort
   let html_lines = s:get_html_template(converted['template_name'])
 
   " processing template variables (refactor to a function)
-  call map(html_lines, 'substitute(v:val, "%title%", "'. converted['title'] .'", "g")')
-  call map(html_lines, 'substitute(v:val, "%date%", "'. converted['date'] .'", "g")')
+  call map(html_lines, 'substitute(v:val, "%title%", converted["title"], "g")')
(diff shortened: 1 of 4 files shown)


chipsenkbeil:
It passes a variety of vim and neovim variants, so I see no reason to not merge this in. @tinmarino, we can obviously keep an eye on recent changes in case something breaks and we need to revert.

tinmarino:
Thank you @brennen, I'm glad you solved it and especially that you put an eye (and a commit) on the tests.
Good job !
@chipsenkbeil thanks for merging.

Such a dirty line it was:
```
call map(html_lines, 'substitute(v:val, "%title%", "'. converted['title'] .'", "g")')
```

See you guys in an other ticket 👋🤓

PR #1128, opened 2021-05-20
VimwikiGoto: Support wiki pages with spaces

sstallion:
This PR updates `:VimwikiGoto` to support opening wiki pages with spaces in the name.

### DIFF
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -3924,6 +3924,7 @@ http://code.google.com/p/vimwiki/issues/list. They may be accessible from
 https://github.com/vimwiki-backup/vimwiki/issues.
 
 New:~
+    * PR #1128: VimwikiGoto: Support wiki pages with spaces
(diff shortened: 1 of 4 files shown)


tinmarino:
Perfect to me:

Just **one byte** to  replace the parameters command from <f-args> by <q-args> so that argument is quoted, anyway only the first argument is considered.

But this one byte is consolidated by nice test.
Thank you @sstallion 

Can someone else please merge (@brennen , @chipsenkbeil)?

chipsenkbeil:
I'm not that familiar with `<q-args>` versus `<f-args>`. Reading through the documentation, it sounds like `<q-args>` will produce a single argument whereas `<f-args>` breaks up by the space. The actual impl of `vimwiki#base#goto` only uses the first argument as part of the full path and an optional second argument as the anchor, so I think the above change makes sense and works better.

Since none of the tests fail, we can give this a go.

Issue #1038, opened 2020-10-21
Feature Request: Only apply links_space_char to __LinkUrl__ part of Markdown link


davidsierradz:
Prior to submitting a new issue make sure to complete these steps:

- [X] Checkout the `dev` branch and confirm the issue is present there as well.
      The `dev` branch contains fixes that may not have been merged to `master` yet.
- [X] Post the syntax you are using (default/mediawiki/markdown) **and** your vimwiki settings from your `.vimrc`
- [X] Provide a detailed description of the problem including **steps to reproduce the issue**.
- [X] Include the output of `:VimwikiShowVersion`.

```
:VimwikiShowVersion
Os: Linux                      
Vim: 800                       
Branch: dev                    
Revision: e2a6b82              
Date: 2020-10-19 23:47:10 -0300
```

Hi. With:

```viml
let g:vimwiki_list = [
      \ {'path': '~/notes/',
      \ 'links_space_char': '-',
      \ 'syntax': 'markdown', 'ext': '.md'}]
```

Currently, if I generate a link, this happens:

```
foo bar -> [foo-bar](foo-bar)
```

I would like:

```
foo bar -> [foo bar](foo-bar)
```

The following patch resolves this:

```diff
diff --git a/autoload/vimwiki/base.vim b/autoload/vimwiki/base.vim
index 5c532d2..3ec366b 100644
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -2719,7 +2719,7 @@ function! s:normalize_link_syntax_v() abort
   let link = s:safesubstitute(link, '__FileExtension__', file_extension , '')
   " Replace space characters
   let sc = vimwiki#vars#get_wikilocal('links_space_char')
-  let link = substitute(link, '\s', sc, 'g')
+  let link = substitute(link, '(\(.*\))', {m -> substitute(m[0], '\s', sc, 'g')}, 'g')
   " Remove newlines
   let link = substitute(link, '\n', '', '')
```

Maybe we can put this behind a config flag? Also I did't tested it with vimwiki syntax, so I don't know if this will break the setting there or cause problems in other parts of the codebase.

Thanks!

tinmarino:
It is a nice idea. But I dont like flags and prefer opcion like descripcion_handler that takes a string like 'substitute(in, '-', ' ', out)' so other peolple can replace say `/` by `@`.

You see the nice thing with that is that we document that is can break auto reference link and push the responsability to the user hihihi.

[...]

PR #1132, opened 2021-06-02
Do not modify spaces to link descriptions

vinnyfuria:
When links_space_char is configure to replace spaces, the spaces should only replace the `__LinkUrl__` and not replace the `__LinkDescription__`

This is a different fix then described in https://github.com/vimwiki/vimwiki/issues/1038 but fixes the described problem.

All vadar tests pass; vint returns no errors

**Note:** on some vim versions I'm seeing 1/4 tests passing in `search.vader`; but that is the case in the `dev` branch as well.

- [X] PRs must pass Vint tests and add new Vader tests as applicable.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -2736,14 +2736,14 @@ function! s:normalize_link_syntax_v() abort
   endif
 
   " Transform link:
-  " Replace description (used for markdown)
-  let link = s:safesubstitute(link, '__LinkDescription__', visual_selection, '')
   " Replace file extension
   let file_extension = vimwiki#vars#get_wikilocal('ext', vimwiki#vars#get_bufferlocal('wiki_nr'))
   let link = s:safesubstitute(link, '__FileExtension__', file_extension , '')
   " Replace space characters
   let sc = vimwiki#vars#get_wikilocal('links_space_char')
   let link = substitute(link, '\s', sc, 'g')
+  " Replace description (used for markdown)
+  let link = s:safesubstitute(link, '__LinkDescription__', visual_selection, '')
   " Remove newlines
   let link = substitute(link, '\n', '', '')
 
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -3974,7 +3974,7 @@ Changed:~
     * VimwikiCheckLinks work on current wiki or on range
     * PR #1057: Change how resolve subdir work
       Change how shell escaping work when using CustomWiki2HTML
-    * PR #TBD: Spaces in link descriptions are no longer replaced with
+    * PR #1132: Spaces in link descriptions are no longer replaced with
       links_space_char
 
 Removed:~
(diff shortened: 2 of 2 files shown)


vinnyfuria:
I'm not sure how to craft a vadar test for this change; assistance or advice for doing so would be appreciated.

tinmarino:
Hi @vinnyfuria,
Tk for the PR, it seems nice from a glimpse and a Vador test would be nice thank you also for proposing it (it is really less job to integrate, maintain).

You can:
1/ Paste here some in / out, I lean formalize the test
2/ Use a private fork of vimwiki and puch on this branch so the jenkins execute.
3/ Run the vimwiki/test/run_test.sh script, it requires linux and docker (even if I tied to windowsify a little)

There is a Readme in the /test, in local Vim use the :Vader command (and Vim plugin) And for the syntax, copy-paste other tests.

In 2h, sure you are done (30 min maybe). But it may be interesting.

vinnyfuria:
Thanks @tinmarino; I'll do my best:

1. To test this change, `links_space_char` should be set to `_` and `syntax` should be `markdown`.
The test input text should be any text that contains spaces, for instance:
```
This is a test
```
Then, in visual mode, the string should be highlighted and then converted to a link by pushing `<enter>`. The result should be:
```
[This is a test](This_is_a_test)
```

2. I'll push an update that includes the PR number in the changelog, hopefully that'll be enough to trigger Jenkins

3. Output from the `run_test.sh` script (forgive the ugliness, I used `script` to capture the output): https://gist.github.com/256e6eb8d9dd280f0a2338c249d78044. As noted in the description, I'm seeing a few test failures in a subset of the versions; but the same test failures are occuring on the `dev` branch without my changes.

vinnyfuria:
I've gotten the outlines of a Vader test:

```
Execute (Log):
  Log 'Markdown with links_space_char set'

Given vimwiki (abc def ghi jkl):
  abc def ghi jkl

Execute (Configure links_space_char for test):
  call SetSyntax('markdown')
  let vimwiki_list_save = g:vimwiki_list
  let g:vimwiki_list = [{'path': '~/wiki/', 'syntax': 'markdown', 'ext': '.md', 'links_space_char': '_'}]

Do (vee<CR>):
  vee\<CR>

Expect (underscores in link url not in description):
  [abc def](abc_def) ghi jkl

Execute (restore):
  let g:vimwiki_list = vimwiki_list_save
```

However, setting the `links_space_char` in `vimwiki_list` doesn't seem to be taking effect (and I can reproduce this behavior in "normal" vim). Is there are command to force the reload of the configuration from the `vimwiki_list`? I don't think such a command exists; so I'm not sure how to integrate this test into Vader without breaking all the other tests.

tinmarino:
Tk @vinnyfuria I just added the tests, It is always a hack to modify vimwiki variables (and restore peace at end).
This is why I did it, after 10 quick and dirty try, the 11th worked.

Thanks for the PR

PR #1133, opened 2021-06-03
Fix self-link for TOC links.

HugoNikanor:
Previously TOC links contained a closing </a>-tag, but lacked the accompanying start tag, leading to invalid HTML. This fixes that, along with placing class="justcenter" on the correct tag once again.


tinmarino:
Markdown
```markdown
        = Header +
```

Before
```html
<div id="Heading"><h1 id="Heading" class="header"><a href="#Heading" class="justcenter">Heading</a></h1></div>
```

After
```html
<div id="Heading"><h1 id="Heading" class="header justcenter"><a href="#Heading">Heading</a></h1></div>
```

PR #1155, opened 2021-07-18
fix(update_listing_in_buffer): don't update listing if there are no c…

tmerse:
…hanges

Before updating a listing, check if the update differs from the existing state.
Only then update the buffer, otherwise return early.

Initial motivation: If `let g:vimwiki_auto_toc = 1` is set, saving a buffer would always update
the `contents`-section, even if there were no updates in this section. This lead to undesired
undo-behavior.

NOTE: this fix was only tested for the toc-listing.

PR. is a WIP, if keeping a clean undo-history qualifies as a problem, I will continue working on this.


tinmarino:
Tk @tmerse,

You added a clause to not work if no change, but in case of changes, the a:Generator.f() is called twice. I added a commit to change that and call the Generator only once.

PR #1158, opened 2021-07-28
Feature: VimwikiPasteLink to get absolute wiki link to the current file

paperbenni:
This is useful for quickly pages that multiple other pages link to. Until now the only two options were to create relative links that are different for different pages and require manual work or to manually type out the absolute path.

### DIFF
--- a/autoload/vimwiki/path.vim
+++ b/autoload/vimwiki/path.vim
@@ -221,6 +221,17 @@ function! vimwiki#path#is_absolute(path) abort
 endfunction
 
 
+
+function! s:get_wikifile_link(wikifile) abort
+  return vimwiki#base#subdir(vimwiki#vars#get_wikilocal('path'), a:wikifile).
+    \ fnamemodify(a:wikifile, ':t:r')
+endfunction
+
+function! vimwiki#path#PasteLink(wikifile) abort
+  execute 'r !echo "[[/'.s:get_wikifile_link(a:wikifile).']]"'
+endfunction
+
+
 if vimwiki#u#is_windows()
   " Combine: a directory and a file into one path, doesn't generate duplicate
   " path separator in case the directory is also having an ending / or \. This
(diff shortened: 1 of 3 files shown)


tinmarino (approved):
Hi @paperbenni,

Thank you for the PR. It is safe to merge from my point.
It adds doc to the undocumented PasteUrl (html path) command and create a coherent new PasteFile command (md file).

Can another reviewer (@brennen or @chipsenkbeil) rebase it please (for the double review).

autoload/vimwiki/path.vim
```diff
@@ -221,6 +221,17 @@ function! vimwiki#path#is_absolute(path) abort
...
 
 
+
+function! s:get_wikifile_link(wikifile) abort
+  return vimwiki#base#subdir(vimwiki#vars#get_wikilocal('path'), a:wikifile).
+    \ fnamemodify(a:wikifile, ':t:r')
+endfunction
+
+function! vimwiki#path#PasteLink(wikifile) abort
+  execute 'r !echo "[[/'.s:get_wikifile_link(a:wikifile).']]"'
```
brennen, line 231: Shelling out to `echo` here seems kind of weird.  I checked, and it looks like it's already happening 3 places, at least:

```
10:53:47 brennen@inertia:~/.vim/bundle/vimwiki (dev *$=) ✰ git grep '!echo'
autoload/vimwiki/html.vim:  execute 'r !echo file://'.s:get_wikifile_url(a:wikifile)
autoload/vimwiki/html.vim:  execute '!echo file://'.s:get_wikifile_url(a:wikifile)
autoload/vimwiki/path.vim:  execute 'r !echo "[[/'.s:get_wikifile_link(a:wikifile).']]"'
```

That should probably be cleaned up.  Could you avoid it here by using `append()` or `insert()` instead?

brennen (changes requested):
Apart from the `echo` issue mentioned above, this looks good to me.

Issue #794, opened 2020-01-02
VimwikiGenerateTags links don't handle file hierarchy correctly


jean:
Create a tag (e.g. `:SomeTag:`), rebuild tags, and in `diary/2020-01-02.wiki` issue the command `:VimwikiGenerateTags`. 

- This results in links like `[[diary/2020-01-02#SomeTag]]`. When that link is followed, vim tries to create `diary/diary/2020-01-02.wiki`. **This is wrong**: 
  - _either_ links should be generated based on the main wiki directory (like `[[/diary/2020-01-02#SomeTag]]`), 
  - _or_ (better) they should be generated relative to the file in which the `:VimwikiGenerateTags` command is issued (`[[2020-01-02#SomeTag]]`). 

This is a follow-on from https://github.com/vimwiki/vimwiki/issues/190.

----

Syntax: default.

## From `.vimrc`:

```vim
let wiki = {}
let wiki.name = 'wiki'
let wiki.path = '~/wiki/'
let wiki.nested_syntaxes = nested_syntaxes
let wiki.auto_tags = 1
```

## The output of `:VimwikiShowVersion`.

Branch: dev
Revision: 5cdeb9a
Date: 2020-01-01 10:58:56 -0700

ranebrown:
Might be able to use `vimwiki#path#relpath(dir, file)` to fix this.

tinmarino:
Fixed #1162

PR #1162, opened 2021-08-18
Fix links generated from VimwikiGenerateTagLinks

RonMcKay:
As described in issue #794 links that are generated from tags are build wrong and are resulting in non functioning links for which vimwiki tries to create new wiki files if followed.

This pull request changes the generated links in `vimwiki#tag#generate_tags` such that they are relative to the file in which the `VimwikiGenerateTagLinks` command is issued.

I am not certain if the tag link should be made relative when iterating over the available tag links (like in the proposed solution) or  earlier in the function when the `tag_entries` dictionary is build. I would be happy if someone who is more knowledgeable would have a look at this.

All vint tests are passing but I do not really know how to write an additional test for this change.
EDIT: I just added a vader test case which tests for tag links relative to the current directory.

### DIFF
--- a/autoload/vimwiki/tags.vim
+++ b/autoload/vimwiki/tags.vim
@@ -345,6 +345,7 @@ function! vimwiki#tags#generate_tags(create, ...) abort
 
     let lines = []
     let bullet = repeat(' ', vimwiki#lst#get_list_margin()).vimwiki#lst#default_symbol().' '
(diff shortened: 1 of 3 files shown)


tinmarino:
Thank you @RonMcKay for the well prepared PR (with tests).
Not sure why Travis tests did not stat automatically but I'll see that after merge.

Issue #621, opened 2019-03-05
syntax highlight in html


jahagirdar:
Is it possible for the exported html to have syntax highlighting for the embedded code fragment?

optik-aper:
Not native to vimwiki, as far as I know.  You could include something like [highlight.js](https://highlightjs.org/), I'd think, but I've never used it.

gpanders:
You can do this as follows:

1. Copy the file `default.tpl` out of `vimwiki/autoload/vimwiki/` into your own directory.
2. Modify your copy of `default.tpl` to add the following lines:

In `<head>`:
```html
<link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.15.6/styles/default.min.css" />
```

At the end of `<body>`:
```html
<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.15.6/highlight.min.js"></script>
<script type="text/javascript">
    document.querySelectorAll('pre').forEach(block => hljs.highlightBlock(block));
</script>
```
Now tell Vimwiki that you want to use this template:
```vim
let g:vimwiki_list = [{'path': '/path/to/wikis', 'template_path': '/path/to/directory/containing/default.tpl'}]
```

Highlight.js will detect syntax automatically. So in your Vimwiki files, you just need to use the `{{{` and `}}}` blocks to designate a `<pre>` section and highlight.js will take care of the rest.

You can also "hint" at the language to help highlight.js by using `{{{class="python"` (or whatever the language you are using is) in your Vimwiki file.

hq6:
I am a new maintainer, and I think this would be nice to have natively. 

However, I will defer to @EinfachToll  and @Nudin until I get a better sense for what flavor of changes they think are appropriate, so you may want to hold off on this for now, while we handle the current PR's.

Nudin:
Hey, I really like to have syntax-highlithing in the HTML-export. But I also think we shouldn't add JS to the default template since most users won't need it. And I think we should handle this and the math-rendering consistently.
Currently for the math rendering we explain in the help-file how to enable it. So we could do the same for this. On the other side I fear that the documentation if becoming to long so that people might stop reading it. I'm planing to create a wiki where to collect tips and tricks – documenting the how-to there and just referencing it from the help-file might be the second option.
[...]

PR #1181, opened 2021-10-29
Addresses issue #621

ysftaha:
See issue #621


tinmarino:
Nice addition @ysftaha , test would have been nice but let's merge that to share the same code.
I'll write the tests when time permits

doc/vimwiki.txt
```diff
@@ -3572,6 +3572,40 @@ The header level of generated links. Valid values are from 1 to 6.
...
+When set to 1, enables syntax highlighting in resulting HTML. Whenever a code
+block is encountered the *g:vimwiki_listing_hl_command* is invoked on the
+code block.
+
+
+Value           Description~
+0               Do not highlight code blocks in HTML.
+1               Highlight code blocks in HTML. see *g:vimwiki_listing_hl_command*
+
+Default: 0
```
lilyinstarlight, line 3587: I think there may have been a typo in this PR's help document that is preventing the help tags from building (I'm getting `154: Duplicate tag "g:vimwiki_listing_hl_command" in file ./doc/vimwiki.txt` [here](https://github.com/lilyinstarlight/foosteros/runs/4571861971?check_suite_focus=true#step:5:599))

I don't really know the help file syntax all that well, but shouldn't references to tags (rather than definitions) use `|` instead of `*` in this section? ([Vim `help-writing` documentation](https://vimhelp.org/helphelp.txt.html#help-writing))

```suggestion
*g:vimwiki_listing_hl*

When set to 1, enables syntax highlighting in resulting HTML. Whenever a code
block is encountered the |g:vimwiki_listing_hl_command| is invoked on the
code block.


Value           Description~
0               Do not highlight code blocks in HTML.
1               Highlight code blocks in HTML. see |g:vimwiki_listing_hl_command|

Default: 0
```

tinmarino: Tk @lilyinstarlight , I've just pushed the fix you sugested and now I can
```
:helptags all
```

Issue #1141, opened 2021-06-13
How to open all wikilinks in current page?


shihabdider:
Is there a way to open all the wikilinks on the current page (e.g as a quickfix list)? This would avoid having to go back and forth between the parent and child pages.

brennen:
Not that I'm aware of.

You can tab between links and open them in a split with Shift + Enter (in some terminals), or bind `:VimwikiSplitLink`, `:VimwikiVSplitLink`, or `:VimwikiTabnewlink` to something, but that doesn't really solve the problem of showing a list of all links in the current page.

I have some custom configuration that would make this pretty easy to do in a script - I'm scraping links in a given page into an sqlite database with this:

  - https://code.p1k3.com/gitea/brennen/bpb-kit/src/branch/main/home/.vim/vimrc#L403
  - https://code.p1k3.com/gitea/brennen/bpb-kit/src/branch/main/home/.vim/vimrc#L848
  - https://code.p1k3.com/gitea/brennen/bpb-kit/src/branch/main/home/bin/notes-collect-metadata

Which lets me run `notes-links-for` to find all the links _to_ a given page:

  - https://code.p1k3.com/gitea/brennen/bpb-kit/src/branch/main/home/bin/notes-links-for

It'd be trivial to find all the ones _in_ a given page the same way, but you need Pandoc, a Python library called Panflute, sqlite, and probably a couple of other things.  I should really document this at some point and package it up for distribution, maybe when I try out using @chipsenkbeil's vimwiki-rs instead of Pandoc for it.

brennen:
(Of course you probably wouldn't need anything as sophisticated as parsing the AST to do this.  It could probably be built into the plugin, but as far as I know there's nothing that does it now.)

PR #1185, opened 2021-11-06
Feature: Add wiki page to buffer list using `badd`

jlbas:
This PR adds the `VimwikiBaddLink` command, which adds wiki pages to the buffer list, if they weren't already listed. The main reason for wanting this feature is to facilitate the process of opening multiple links. For example, using this command, one can open multiple wiki pages into hidden buffers, and then cycle or navigate through them however they like. When opening links, this avoids having to go back and forth or worrying about window layouts.

There has previously been interest for a feature that avoids such back and forth behavior when opening multiple links #1141. However, this PR only partially addresses that issue, since the user wanted a way of opening/listing all the links in the current page (something for a future PR if this fits into the overall design of Vimwiki).


tinmarino:
Thanks @jlbas

Issue #966, opened 2020-08-01
Change default link name in Generated tags


CaptainKryuk:
Prior to submitting a new issue make sure to complete these steps:

- .vimsrc vimwiki settings
![image](https://user-images.githubusercontent.com/35116092/89101567-baae4a00-d409-11ea-8f65-fadfda8c8360.png)

- ![image](https://user-images.githubusercontent.com/35116092/89101527-44114c80-d409-11ea-8912-738df65f08e5.png) -
 my generated tags. But I want to create for example:

"# Generated tags

"## tag_name
" - link1_name
" - link2_name

vimwiki version
- [![image](https://user-images.githubusercontent.com/35116092/89101497-0b717300-d409-11ea-9999-a14f839dc0bd.png)]

tinmarino:
Estimated time: 3h  <= this is a standalone feature, so new commers may not have to dive a lot in the code

Issue #724, opened 2019-06-20
Generate an index for a particular tag


evantravers:
This is just an idea… I sometimes manage "hub" pages in my wiki, and it'd be nice to be able to have a live updating list of pages with just specific tags rather than all tags… something like

```
# Generated Tag :foobar:
```

And that would pull only that tag… sorry if this is a duplicate, I didn't see one.

evantravers:
I'm an idiot. `:VimwikiGenerateTags` does take arguments. For those who find this via google…  it seems that on the `dev` branch, the `auto_tags` and `auto_generate_tags` options will cause them to auto update. Pretty close to ideal for my uses.

evantravers:
Well… `auto_generate_tags` causes the magic header for a single set of tags that you wish to track to automatically expand to _all_ tags.

We may want a magic header like the one I described in the first comment.

ranebrown:
I agree that could be useful. Not sure when anyone will have time to look into it though. If you want to create a PR that would definitely speed things up.

PR #1196, opened 2021-12-19
Improve generation of tag links

RonMcKay:
This PR introduces the following changes to the tag system:

# Collect a link description when building the tag metadata file

Before, the link caption of a generated tag link was set as the anchor of the link, which resulted in ugly looking lists. E.g. the 
following:
```
# This is a header
:this-is-a-tag:
```
Resulted in the generated tag links:
```
# Generated Tags

## this-is-a-tag
- [this-is-a-header](#this-is-a-header)
```
The link caption would look even more strange when the header had a higher level, which results in all parent headers being included in the link.

This was updated to show the real header text before it was formatted as an anchor (a53d605) which results in:

```
# Generated Tags

## this-is-a-tag
- [This is a header](#this-is-a-header)
```

**Attention**: This changes the creation of the meta-data file and thus marks old meta data file versions as corrupted. This can however easily be fixed by deleting the `.vimwiki-tags` metadata file and rebuilding it with `:VimwikiRebuildTags`.

To allow the vader tests to pass, some of the tests in `tag.vader` had to be adjusted for the new header captions and the changed metadata file format. (402ff6f)

# Look for already existing tag links

When setting `vimwiki-option-auto_generate_tags` to `1` and writing a file where `VimwikiGenerateTagLinks` was invoked with only a single tag e.g.. The function `vimwiki#tags#generate_tags` would overwrite the auto generated links for this single tag with the links to all tags available in the current wiki. This does not seem to be the intent of the `auto_generate_tags` option.

The function was changed so that when having the option enabled the buffer will now be searched for the `vimwiki#vars#get_global('tags_header')` (`Generated Tags` by default) and all tags that were used to build the auto generated links will be used to update these instead of overwriting it with all available tags in the wiki. (64c606b)

For this to work I made the function `s:collect_headers` a plugin wide function (so I changed it to `vimwiki#base#collect_headers`) as it was defined in `base.vim`. This way one can use it in `tags.vim` to retrieve all headers in the file and iterate over them. (b1deba9) Obviously I also changed all occurences of `collect_headers()` in `base.vim` to `vimwiki#base#collect_headers()`.

Additionally I added another vader test case to cover the updating of the auto generated link section. (b079a47)

All vader and vint tests are passing for me.

----

Adresses #724
Might be related to #966 but I could be wrong as the information there is very limited.

I have not yet updated the changelog as I am not sure if that is fixing #724.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1968,7 +1968,7 @@ endfunction
 
 function! vimwiki#base#TO_header(inner, including_subheaders, count) abort
   " Jump to next header (Exported for text object)
-  let headers = s:collect_headers()
+  let headers = vimwiki#base#collect_headers()
   if empty(headers)
     return
   endif
@@ -2275,7 +2275,7 @@ function! vimwiki#base#RemoveHeaderLevel(...) abort
 endfunction
 
 
-function! s:collect_headers() abort
+function! vimwiki#base#collect_headers() abort
   " Returns: all the headers in the current buffer as a list of the form
   " [[line_number, header_level, header_text], [...], [...], ...]
   " Init loop variables
@@ -2392,7 +2392,7 @@ endfunction
 
 function! vimwiki#base#goto_parent_header() abort
   " Jump to parent header
-  let headers = s:collect_headers()
+  let headers = vimwiki#base#collect_headers()
   let current_header_index = s:current_header(headers, line('.'))
   let parent_header = s:get_another_header(headers, current_header_index, -1, '<')
   if parent_header >= 0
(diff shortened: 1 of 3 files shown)


RonMcKay:
As a small update: I am using this for a few months now quite actively and I did not have found any bugs yet.

tinmarino:
Perfect PR, I did not even have to test it on my computer (just review on phone).
* Other feature not tuched
* Test included
* Comments in code
* Good description in this thread

Good job to ping again: as no-one complained, we just forgot about it (at least from my part).

@sdondley next time I'll comment a "go" so that a peer can merge: it worked fast last time. Please do the same if you see no problem.

PR #1211, opened 2022-02-16
Support angle bracket escaped markdown urls

CharlesSchimmel:
Very small change to support Markdown links with angle bracket escaped URIs, like `[link title](<path/to/some file>)` to enable link-following of angle bracket escaped markdown links. This syntax is [part of the the Commonmark spec](https://spec.commonmark.org/0.30/#example-487):

> The [link] destination can only contain spaces if it is enclosed in pointy brackets

 This is also helpful is people use Obsidian alongside their Vimwiki as Obsidian strongly conforms to markdown standards and will not parse links with enescaped spaces. I believe this is all needed to support link following and concealment/preview.

Edited to specifically highlight link-following of angle bracket escaped markdown links and add related commonmark spec

### DIFF
--- a/autoload/vimwiki/vars.vim
+++ b/autoload/vimwiki/vars.vim
@@ -1181,7 +1181,7 @@ function! s:populate_extra_markdown_vars() abort
   let mkd_syntax.rxWeblinkMatchDescr0 = mkd_syntax.rxWeblinkMatchDescr
 
   let mkd_syntax.rxWeblink1Prefix = '['
-  let mkd_syntax.rxWeblink1Suffix = '>\=)'
+  let mkd_syntax.rxWeblink1Suffix = ')'
   let mkd_syntax.rxWeblink1EscapeCharsSuffix = '\(\\\)\@<!\(>\=)\)'
   let mkd_syntax.rxWeblink1Separator = ']('
 
--- a/autoload/vimwiki/vars.vim
+++ b/autoload/vimwiki/vars.vim
@@ -1181,8 +1181,8 @@ function! s:populate_extra_markdown_vars() abort
   let mkd_syntax.rxWeblinkMatchDescr0 = mkd_syntax.rxWeblinkMatchDescr
 
   let mkd_syntax.rxWeblink1Prefix = '['
-  let mkd_syntax.rxWeblink1Suffix = ')'
-  let mkd_syntax.rxWeblink1EscapeCharsSuffix = '\(\\\)\@<!\()\)'
+  let mkd_syntax.rxWeblink1Suffix = '>\=)'
+  let mkd_syntax.rxWeblink1EscapeCharsSuffix = '\(\\\)\@<!\(>\=)\)'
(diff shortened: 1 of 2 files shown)


autoload/vimwiki/vars.vim
```diff
@@ -1207,7 +1207,7 @@ function! s:populate_extra_markdown_vars() abort
   let valid_chars_url = '[^[:cntrl:]]'
 
   let mkd_syntax.rxWeblink1Prefix = vimwiki#u#escape(mkd_syntax.rxWeblink1Prefix)
-  let mkd_syntax.rxWeblink1Separator = vimwiki#u#escape(mkd_syntax.rxWeblink1Separator)
```
tinmarino, line 1210: This is hardcoded => bad
Please better change the default data in vars.vim

tinmarino (commented):
Please copy-paste a snippet of what you see before and after the  changes.
Also note the hardcoded regex because this may break some user configuration.
A link to that markdown spec in comment would be nice (easier to verify).

That said, the improvement seems nice, thank you for the effort. With small changes to get it safe, sure it will be accetpted.

CharlesSchimmel:
Visually there will be no change, markdown-style links will be concealed the same way. I should have mentioned that this allows `vimwiki#base#follow_link()` to work with angle bracket escaped markdown links without trying to include the angle brackets in the file name (navigating to `foo` in `[foo](<foo>)` instead of trying to navigate to `<foo>`).

[Commonmark](https://spec.commonmark.org/0.30/#example-487):
> The destination can only contain spaces if it is enclosed in pointy brackets

Pandoc doesn't mention it in the docs but [it's in the code](https://github.com/jgm/pandoc/blob/dc534caf7a16e5f4456e6a7676024ab1fc86b3ee/src/Text/Pandoc/Readers/Markdown.hs#L1819) if that helps.

Thanks! I will fix up the regex shortly.

autoload/vimwiki/vars.vim
CharlesSchimmel: Would something like `let mkd_syntax.rxWeblink1Separator = vimwiki#u#escape(mkd_syntax.rxWeblink1Separator).'<\='` be preferable or should I pull this into a new variable entirely? I'm not sure how to escape it without ruining the `\=`

hulufei:
@CharlesSchimmel @tinmarino this is exactly what I want. Anything I can help with? I'd like to get this PR merged.

zaldih:
I was coming to check if just this feat had been discussed before to work on it and make a PR. I'm glad to see that someone has already worked on it.

I would like to see this PR merge!

tinmarino:
Rebased with squashed <= It works fine actually

I am really sorry for the delay, got confused => Was waiting for @CharlesSchimmel  who was actually waiting for my answer. Sorry guys.

And well thank you very much @CharlesSchimmel for the PR (and the patience). I hope to see you in another PR 😃

PR #1234, opened 2022-05-22
Reuse existing tabs with tab drop in following links and opening diary pages (closes #238)

### LINKED ISSUE #238: Enhancement: Reuse Vim Tabs with drop
I use vim and gvim opening many files in new tabs. I found that when following links forward and back vimwiki would frequently create a new tab for a file that was already open, so that I would end up with the same file open many times. My preference is to reuse the existing tab.

Below are two patch files for how I hacked my local vimwiki to reuse tabs instead of creating new ones.

```

--- markdown_base.vim   2016-07-02 14:13:32.750195233 -0700
+++ markdown_base.vim   2016-07-02 14:11:47.274198885 -0700
@@ -86,9 +86,9 @@
     elseif a:split ==# "vsplit"
       let cmd = ":vsplit "
     elseif a:split ==# "tabnew"
-      let cmd = ":tabnew "
+      let cmd = ":tab drop "
     else
-      let cmd = ":e "
+      let cmd = ":drop "
     endif

     " try WikiLink

```

```

--- base.vim    2016-07-02 14:13:21.342195628 -0700
+++ base.vim    2016-07-02 14:08:44.654205208 -0700
@@ -1270,7 +1270,7 @@
   if exists("b:vimwiki_prev_link")
     " go back to saved wiki link
     let prev_word = b:vimwiki_prev_link
-    execute ":e ".substitute(prev_word[0], '\s', '\\\0', 'g')
+    execute ":drop ".substitute(prev_word[0], '\s', '\\\0', 'g')
     call setpos('.', prev_word[1])
   else
     " maybe we came here by jumping to a tag -> pop from the tag stack

```

brennen:
Replaces :e in `vimwiki#base#follow_link()` with `:drop`, making this the default behavior for pressing `<CR>` on a link.

Adds a new `:VimwikiTabDropLink` and makes this a default for the keybindings formerly occupied by `:VimwikiTabnewLink`; leaving `:VimwikiTabnewLink` available for backwards compatibility and anyone who still wants the old behavior.

Also defaults to using `:drop` for `:VimwikiMakeDiaryNote`, eliminating another common source of duplicate tabs / buffers.

Doesn't touch the split window reuse functionality, or the :VimwikiGoBackLink behavior, although I can see an argument for adding :drop to the latter.

I've wanted these changes for a while and happened to notice @davidlmontgomery's patch from 2016 in #238.  Code has moved around a little since, but I think this is correct.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1664,8 +1664,17 @@ function! vimwiki#base#follow_link(split, ...) abort
       let cmd = ':badd '
     elseif a:split ==# 'tab'
       let cmd = ':tabnew '
+    elseif a:split ==# 'tabdrop'
+      " Use tab drop if we've already got the file open in an existing tab
+      let cmd = ':tab drop '
     else
+      " Same as above - doing this by default reduces incidence of multiple
+      " tabs with the same file.  We default to :e just in case :drop doesn't
+      " exist in the current build.
       let cmd = ':e '
+      if exists(':drop')
+        let cmd = ':drop '
+      endif
     endif
 
     " if we want to and can reuse a split window, jump to that window and open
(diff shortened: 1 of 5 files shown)


brennen:
Self-merging after a couple months of silence here.  I've been using this steadily since May and it seems to work without issue.

PR #1240, opened 2022-06-10
make vimwiki#html#Wiki2HTML respect 'html_filename_parameterization'

jsn:
Fixes `:Vimwiki2HTMLBrowse` when vimwiki_markdown is in use.


brennen:
Under manual testing, this seems to work. Code looks reasonable.

Issue #1244, opened 2022-06-22
VimwikiBacklinks doesn't see wiki-style [[links]] when using markdown syntax


nebulaeandstars:
tl;dr: VimwikiBacklinks can recognise \[normal-links\](normal-links), but not [[wiki-style-links]]. Given that vimwiki seems to support both formats, I assume this is a bug.

- expected behaviour: both forms of link should be counted as a backlink.
- actual behaviour: only \[markdown-style links\](some-other-note) are detected.
- bug is present in the latest versions of both `dev` and `master`.

vimwiki config:
```vim
let g:vimwiki_list = [{
            \'path': '~/docs/notes/',
            \'path_html': '~/docs/notes/html',
            \'syntax': 'markdown',
            \'ext': '.md',
            \}]

let g:vimwiki_table_mappings = 0
let g:vimwiki_global_ext = 0
```

Steps to reproduce:
1. Make three files
2. Add two links to the first file:
    - `[second-file](second-file)`
    - `[[third-file]]`
3. Follow the links and test using `:VWB`
    - the second file will have a backlink to the first
    - the third file will not

nebulaeandstars:
I'm not that familiar with vimscript, but I've made an MR for this, as it ended up being a fairly quick fix

fcsm1922:
@nebulaeandstars,
> I'm not that familiar with vimscript, but I've made an MR for this, as it ended up being a fairly quick fix

Hi.Please tell me how you solved this problem?

nebulaeandstars:
I just changed part of the function that generates the list of links.

Before it was saying "If in markdown mode, count markdown links, otherwise count wiki links."

Now it says "Count wiki links. If in markdown mode, also count markdown links."

Pull request is [here](https://github.com/vimwiki/vimwiki/pull/1245). It doesn't like it can merge anymore, though.

fcsm1922:
@[nebulaeandstars](https://github.com/nebulaeandstars), 

did you change this in the base.vim file?
could you post your file? so as not to make mistakes when changing))

fcsm1922:
made changes to the file ~/.vim/bundle/vimwiki/autoload/vimwiki/base.vim that you made.... and nothing..
and yes... the base.vim file is smaller for some reason (by the number of lines of code)

in the created vimwiki directory
[[folder/filename]]
when converting index.md to html file index.html outputs
[[folder/filename]] instead of a link to a file located in folder

VimwikiShowVersion: stable 2.5

[...]

PR #1245, opened 2022-06-22
fix wiki-style backlinks for markdown syntax

nebulaeandstars:
Very small bugfix - addresses #1244 

Changes the way s:get_links finds/lists links within a file, so that both \[regular-links\](regular-links) and [[wiki-links]] are recognised when using markdown syntax. This should make VimwikiBacklinks line up with everything else.


brennen:
Tested, seems to work as expected.

Merged as 3ac8e1a with contributor list update & changelog entry. Apologies for not setting `Author` there correctly, but I'll spare everyone the potential hassle of a force push with a rewrite.

PR #1282, opened 2022-11-30
Prepare 2.6.0 release

brennen:
This:

  - Merges in a couple of old changes from `master`
  - Changes version number everywhere I could find it.
  - Reworks get_version() to dispense with stable / dev division.
  - Rewrites CONTRIBUTING.md to describe a new release workflow.

The development/release workflow stuff could use some thought.  It's clunky at best.  There's sort of an inherent tension between semver-style version numbers and the rolling release thing - maybe we should just use commit hash or something date-based?

Still, it's a start.

As a followup, I tested an action for mirroring `dev` to `master` using [google/mirror-branch-action](https://github.com/google/mirror-branch-action). I'm not super familiar with GitHub Actions, but that seems to work.

```yaml
cat .github/workflows/mirror.yml 
on:
  push:
    branches:
      - 'dev'

jobs:
  mirror_job:
    runs-on: ubuntu-latest
    name: Mirror SOURCE_BRANCH_NAME branch to DESTINATION_BRANCH_NAME branch
    steps:
    - name: Mirror action step
      id: mirror
      uses: google/mirror-branch-action@v1.0
      with:
        github-token: ${{ secrets.GITHUB_TOKEN }}
        source: 'dev'
        dest: 'master'
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,10 +10,12 @@ Issues can be filed at https://github.com/vimwiki/vimwiki/issues/
 
 # Git branching model
 
-As of v2.6.0, VimWiki has adopted a rolling release model, while (for the
-moment) retaining specific version numbers.  A release should be
+As of v2022.12.02, VimWiki has adopted a rolling release model, along with
+[calendar versioning][calver].  A release should be
 [prepared][#preparing-a-release] for every change or set of changes which merge
 to `dev`.
+
+[calver]: https://calver.org/
 
 There are two permanent branches:
     1. `dev`: This is the default branch, and where changes are released. Tasks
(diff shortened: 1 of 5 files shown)


jamestomasino:
> maybe we should just use commit hash or something date-based?

https://calver.org/ is an excellent option here. I'd recommend something like `YYYY.MM.DD_MICRO` used by [fusefs-ntfs](http://www.freshports.org/sysutils/fusefs-ntfs).

Better than hashes, it's easy to tell when you're out of date and by how much. It also gives the optional micro descriptor for technical edge-cases.

brennen:
Yeah, I like the date-based idea.  It at least still carries some info.  Still clunky updating a version with each change, but we can probably streamline things over time.

Issue #1297, opened 2023-01-19
Table formatting breaks on multibyte separator


druckdev:
Prior to submitting a new issue make sure to complete these steps:

- [X] Include the VimWiki settings from your `.vimrc`:
```
if (!exists('g:vimwiki_syntax_list'))                     
	let g:vimwiki_syntax_list = {}                        
endif                                                     
if (!has_key(g:vimwiki_syntax_list, 'markdown'))          
	let g:vimwiki_syntax_list['markdown'] = {}            
endif                                                     
let g:vimwiki_syntax_list['markdown']['rxTableSep'] = '│' 
```
- [X] Include the syntax you are using (~default~ / Markdown / ~MediaWiki~) 
- [X] Include the output of `:VimwikiShowVersion`:
```
Version: 2022.12.02
Os: Linux
Vim: 800
Branch: dev
Revision: fea8bee382b2
Date: 2022-12-04 21:25:52 -0700
```
- [X] Provide a detailed description of the problem including **steps to reproduce the issue**:

## Description

When setting a multibyte table separator (e.g. vertical box drawing character / Unicode U+2502), the table formatting breaks and clears the whole table.

## Steps to reproduce

1. `:VimwikiTable`
2. Move the cursor into the new table (which is generated with the correct separator)
3. Enter insert mode
4. Exit insert mode

tinmarino:
Closing <= Fixed by PR #1298

PR #1298, opened 2023-01-19
Fix: Handle multibyte chars properly in get_cells

druckdev:
Previously `vimwiki#tbl#get_cells()` would act on `line` as a normal string (i.e. using `[idx]` and `strpart()`). This breaks when the column separator (`s:s_rep()` or `rxTableSep`) is a multibyte character like the vertical box drawing character `│`. In that case the separator is never found and only one empty cell is returned. This also affects `vimwiki#tbl#format()` which in turn clears the whole table each time the insert mode is left.

Fix this issue by initially splitting the line into a list of the characters. Getting the elements with `[idx]` now returns each character instead of a byte. In addition `len()` is used in place of `strlen()` as well as slicing together with join() instead of `strpart()`.

Fixes: #1297 ("Table formatting breaks on multibyte separator")

### DIFF
--- a/autoload/vimwiki/tbl.vim
+++ b/autoload/vimwiki/tbl.vim
@@ -149,7 +149,9 @@ function! vimwiki#tbl#get_cells(line, ...) abort
   let state = 'NONE'
   let cell_start = 0
   let quote_start = 0
-  let len = strlen(a:line) - 1
+  " Split byte string into list of character to properly handle multibyte chars
+  let chars = split(a:line, '\zs')
+  let len = len(chars) - 1
 
   " 'Simple' FSM
   while state !=# 'CELL'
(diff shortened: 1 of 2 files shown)


druckdev:
CONTRIBUTING.md states:

> 1. Update the changelog to include, at the top of it, information on the new
   feature the PR introduces or the bug it is fixing as well as the PR number
   and related issue number if possible.

I was unsure on where to place the bugfix note as the current changelog only goes to the current release.  How should I proceed?

druckdev:
Also: I'd be interested in writing test cases for this. But I am unsure how I would change the separator.
Without the lines from the issue in my vimrc, I am not able to change the separator at runtime with following steps:
1. `let g:vimwiki_syntax_list = {'markdown': {'rxTableSep': '│'}}`
2. `unlet g:vimwiki_syntaxlocal_vars`
3. `call vimwiki#vars#init()`

Issue #1229, opened 2022-05-09
Bug: VimwikiGoto without argument creates empty page


rddunphy:
Prior to submitting a new issue make sure to complete these steps:

- [x] Checkout the `dev` branch and confirm the issue is present there as well.

This only affects `dev` branch - on `master` the behaviour is still the same as prior to #913.

- [x] Post the syntax you are using (default/mediawiki/markdown) **and** your vimwiki settings from your `.vimrc`

Using the vimwiki syntax with the following settings in `.vimrc`:
```vim
let g:vimwiki_ext2syntax = {}
let g:vimwiki_table_mappings = 0
let g:vimwiki_use_calendar = 1
let g:vimwiki_auto_diary_index = 1
```
- [x] Provide a detailed description of the problem including **steps to reproduce the issue**.

On the `dev` branch, calling `:VimwikiGoto` without specifying the page name causes a new page just called `.wiki` to be opened. The same happens with the `<leader>wn` global mapping. Instead, the "Enter name" prompt should be shown, as happens on the `master` branch.

- [x] Include the output of `:VimwikiShowVersion`.

Os: Linux
Vim: 801
Branch: dev
Revision: 61c513d
Date: 2022-04-09 20:14:04 -0400

rddunphy:
That works now, thank you! Hitting `<esc>` while in the `Enter name:` prompt still creates the `.wiki` page, whereas I would have expected it to just take me out of the prompt - this is not such a big deal, though.

I did notice another issue while testing this: tab completing a file name with a space in it incorrectly escapes the space (e.g. `<leader>wnso<tab>` results in `Enter name: some\ page`, which rather than taking you to `some page` creates a new file with a backslash in the name). I can open a new issue for this if that would be helpful.

tinmarino:
Hi @rddunphy,

I can reproduce your second bug creating  'file\ with\ space.md'.
Let's keep that second wave on this issue.

tinmarino:
Closed: Last wave fixed by 738e24134ee02c6

PR #1324, opened 2023-04-02
File name prompt fixes

rddunphy:
Fixes two issues mentioned in #1229

In the `Enter name:` prompt opened by `<leader>wn`:
- `<esc>` now just takes you out of the prompt without creating an empty `.wiki` page
- special characters are no longer escaped, since they were being escaped twice (once in `vimwiki#base#get_globlinks_escaped()` and once in `vimwiki#base#edit_file()`)


tinmarino:
Merging <= works fine, no side effect.

PR #1340, opened 2023-05-12
Fix vimwiki#diary#calendar_sign when g:vimwiki_list is not set

### LINKED ISSUE #1336: vimwiki and calendar-vim
" Hook for calendar.vim
if vimwiki#vars#get_global('use_calendar')
  let g:calendar_action = 'vimwiki#diary#calendar_action'
  " This conflicts with calendar.vim plugin
  " let g:calendar_sign = 'vimwiki#diary#calendar_sign'
endif

calendar-vim gives me lots of errors if I do not comment:
" let g:calendar_sign = 'vimwiki#diary#calendar_sign'

Could you check this?
I think, if it is not a bug, it is a configuration clash...

v

LukeDAtkinson:
If a user is happy with the default vimwiki location, they may not set the g:vimwiki_list global variable. In this case, the len(g:vimwiki_list) vimwiki#diary#calendar_sign function throws an error every time it is called. This function is called for every day displayed in the calendar which is very noisy. Checking if the variable exists before doing the length check prevents the error and the rest of the function works fine if the variable is not set.

This fixes #1336 , removing the need for the workaround I identified in that issue's discussion

### DIFF
--- a/autoload/vimwiki/diary.vim
+++ b/autoload/vimwiki/diary.vim
@@ -523,7 +523,7 @@ endfunction
 function! vimwiki#diary#calendar_sign(day, month, year) abort
   " Callback function for Calendar.vim
   " Clause: no wiki no sign #290
-  if len(g:vimwiki_list) <= 0
+  if exists('g:vimwiki_list') && len(g:vimwiki_list) <= 0
     return
   endif
   let day = s:prefix_zero(a:day)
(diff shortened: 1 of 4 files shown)


tinmarino:
Hi @LukeDAtkinson, Thank you for the defensive programming addition.

Accepted <= safe bug fix.

PR #1341, opened 2023-05-13
Fix error when opening links using system handler on Windows

Caid11:
Under Windows and Vim 9, opening a link that triggers the system handler fails with a "command not found" error message. It looks like the title argument to `start` is treated as the executable path.

According to the help page for `start`, the title argument is optional, and /B instructs start not to open a CMD window, so the title is never used and can be removed with no loss in functionality.

This change removes the redundant title argument, fixing the "command not found" error.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -313,7 +313,7 @@ function! vimwiki#base#system_open_link(url) abort
       else
         let url = shellescape(a:url, 1)
       endif
-      execute 'silent ! start "Title" /B ' . url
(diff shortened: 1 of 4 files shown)


Caid11:
After more investigation, I believe this is a bug in Vim itself. Outside of Vim, the title argument is required for `start` to handle web URLs properly, and `start "Title" /B ...` works properly. Furthermore, vim has special handling for start (see `:help !start`). Given all of that, I suspect Vim has a bug with its `start` handling.

Closing this PR to avoid breaking URL handling.

PR #1353, opened 2023-07-04
Make auto_tags preserve existing tag section structure

jiamingc:
Currently, auto_tags will delete sections from "Generated Tags" for tags that aren't used in any pages. This PR makes auto_tags preserve the subsections for unused tags, making auto_tags more usable for tags that are regularly removed (for example a "todo" tag).

Additionally, I made auto_tags preserve the existing order of subsections under "Generated Tags" instead of alphabetizing it.

somewhat related issue: #724

### DIFF
--- a/autoload/vimwiki/tags.vim
+++ b/autoload/vimwiki/tags.vim
@@ -360,65 +360,70 @@ function! vimwiki#tags#generate_tags(create, ...) abort
 
     " make a dictionary { tag_name: [tag_links, ...] }
     let tags_entries = {}
+    for tagname in self.specific_tags
+      let tags_entries[tagname] = []
+    endfor
+
     for entries in values(metadata)
       for entry in entries
         if has_key(tags_entries, entry.tagname)
           call add(tags_entries[entry.tagname], [entry.link, entry.description])
         else
-          let tags_entries[entry.tagname] = [[entry.link, entry.description]]
+          if need_all_tags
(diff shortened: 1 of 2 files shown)


tinmarino:
Hi @jiamingc,

Thank you for the PR, I like the idea to preserve the previous Generated tags structure (otherwise the "create" boolean is kind of useless).

I understand that you can trigger your feature calling `call vimwiki#tags#generate_tags(0)`.

But in that case, the new tags are not appended either, the tag headers remains the same.

It seems that you new feature is safe (conservative) in case of tag removal but do not help in case of tag appending.

Is that the desired behaviour? Otherwise, could you update it so that it does not remove any header (tag name) but yes it appends the new tags in user's wiki.

jiamingc:
Hi, thanks for taking a look at my PR. Not adding new tags to an existing "Generated Tags" section is the current behavior on dev:

https://github.com/vimwiki/vimwiki/blob/65b9f0e68b1c4b6c2023a02b22d11a178e13445e/autoload/vimwiki/tags.vim#L378

This is also the desired behavior to me, since I'm using auto_tags for fixed sets of tags only.

tinmarino:
Hi @jiamingc, Thank you for the quick and clear answer.

Accepted <= do not change existing behaviour and adds feature usefull for many and in vimwiki scope (i.e. hard to put in vimrc).

Issue #1392, opened 2024-02-06
Link works for VIM 9.0 but report error for VIM 9.1


ericsunplus:
Prior to submitting a new issue make sure to complete these steps:

- [√] Include the VimWiki settings from your `.vimrc`
```vim
" Wiki Directories and Settings
let wiki_1 = {}
let wiki_1.path = 'd:/gitlab/mywiki/projDiary/Content'
let wiki_1.path_html = 'd:/gitlab/mywiki/projDiary/Html'
let wiki_1.diary_sort = 'asc'
let wiki_1.nested_syntaxes = {'py': 'python', 'c': 'c', 'bat': 'dosbatch'}
let wiki_1.links_space_char = '_'
let wiki_1.diary_caption_level = -1
let wiki_1.diary_frequency = 'weekly'
let wiki_1.diary_start_week_day = 'wednesday'
let wiki_2 = {}
let wiki_2.path = 'd:/gitlab/mywiki/test/'
let wiki_2.path_html = 'd:/gitlab/mywiki/test_html/'
let wiki_2.diary_sort = 'asc'
let wiki_2.nested_syntaxes = {'py': 'python', 'c': 'c', 'bat': 'dosbatch'}
let wiki_2.diary_frequency = 'weekly'
let wiki_2.diary_start_week_day = 'wednesday'
let g:vimwiki_list = [wiki_1, wiki_2]
let g:vimwiki_conceal_pre = 1
let g:vimwiki_valid_html_tags = 'b,i,s,u,sub,sup,kbd,br,hr,span'
" Don't have menu
let g:vimwiki_menu = ''
" Integrate with Calendar
let g:vimwiki_use_calendar = 1
```
- [√] Include the syntax you are using (default / Markdown / MediaWiki) 
  `default`
- [√] Provide a detailed description of the problem including **steps to reproduce the issue**.
  Press `Enter` on a link works on V9.0 but report the following error on V9.1 vim in windows 10.
![image](https://github.com/vimwiki/vimwiki/assets/7379534/e4130f65-9850-4f1a-9f09-19b5e62bef80)
Looking through `vimwiki\autoload\vimwiki\base.vim` it seems `s:win32_handler` defined inside `vimwiki#base#system_open_link` is not recognized.

If I install back the V9.0 VIM version the issue is gone.

- [√] Include the output of `:VimwikiShowVersion`.
[...]

Issue #1343, opened 2023-05-18
Following HTML link is broken if shell is PowerShell


avysk:
Prior to submitting a new issue make sure to complete these steps:

```
Plug 'vimwiki/vimwiki'
let g:vimwiki_list = [
      \ {'path': '~/OneDrive/vimwiki', 'list_margin': 2},
      \ {'path': '~/vimwiki', 'list_margin': 2} ]

let g:vimwiki_ext2syntax = {}
let g:vimwiki_folding = 'syntax'

autocmd FileType vimwiki setlocal tw=80
autocmd FileType vimwiki setlocal nowrap
autocmd FileType vimwiki setlocal foldmethod=syntax
autocmd FileType vimwiki setlocal foldlevel=2

```

Mediawiki. I am using Windows / PowerShell (core) and my `shell` is set to `pwsh`. Any attempt to open HTML link ends in `E371: Command not found`. This is not very surprising since in `autoload/vimwiki/base.vim` in `win32_handler` it tries to call `start` which does not exist. Replacing `start "Title" /B` with `pwsh -nop -nol -c start` in line 328 fixes the problem.

Version 2023.04.04.

tinmarino:
Hi @avysk,

Thank you for the report and the fix.
Could you please create a PR working on both PowerShell and Batch ?

FocusedWolf:
I'm having an issue very similar with HTML links, E371: Command not found.

For me the problem is on line: 316

Changing: execute 'silent ! start "Title" /B ' . url
To:            execute 'silent ! start ' . url
fixes it.

This might be a vim bug because the code looks correct.

I started gvim without any vimrc using: gvim -u NONE
and doing
:! start "Title" /B https://www.google.com
:! start "" /b https://www.google.com
:! start "" https://www.google.com
all give "E371: Command not found. shell returned -1"
:! start https://www.google.com
works.

Also no issue if i open cmd.exe and do [$ start "" /b https://www.google.com]. Weird.

jfishe:
> I'm having an issue very similar with HTML links, E371: Command not found.
> 
> For me the problem is on line: 316
> 
> Changing: execute 'silent ! start "Title" /B ' . url To: execute 'silent ! start ' . url fixes it.

The problem appears to be single quotes in the `url` string.

```vim
set shellslash
let url = shellescape('https://github.com/vimwiki/vimwiki/issues/1343#issuecomment-1690936595',1)
let url
```

Note the quotes surrounding the string.

```text
[...]

PR #1396, opened 2024-03-08
Issue1343 system_open_link E371 command not found Windows PowerShell and cmd

jfishe:

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -262,11 +262,20 @@ function! vimwiki#base#resolve_link(link_text, ...) abort
               \ vimwiki#vars#get_wikilocal('ext', link_infos.index)
       endif
     else
-      " append extension iff one not already present or it's not the targeted
+      " append extension if one not already present or it's not the targeted
       " wiki extension - https://github.com/vimwiki/vimwiki/issues/950
       let ext = fnamemodify(link_text, ':e')
       let ext_with_dot = '.' . ext
-      if ext ==? '' || ext_with_dot !=? vimwiki#vars#get_wikilocal('ext', link_infos.index)
+
+      " Check if a .md must be added
+      " See #1271 to modify files with a "."
+      let do_add_ext = ext ==? ''
+      if vimwiki#vars#get_syntaxlocal('open_link_add_ext')
+        let do_add_ext = do_add_ext || ext_with_dot !=? vimwiki#vars#get_wikilocal('ext', link_infos.index)
+      endif
+
+      " Add the dot
+      if do_add_ext
         let link_infos.filename .= vimwiki#vars#get_wikilocal('ext', link_infos.index)
       endif
     endif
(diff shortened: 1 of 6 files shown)


autoload/vimwiki/base.vim
```diff
@@ -313,7 +313,7 @@ function! vimwiki#base#system_open_link(url) abort
       else
         let url = shellescape(a:url, 1)
       endif
-      execute 'silent ! start "Title" /B ' . url
+      execute 'silent !start ' .. url
```
tinmarino, line 316: I see that the `..` operator is synonym of the `.`
It would be nice to keep the same style for all Vimwiki code hence using the single dot `.`.

Please replace `..` by `.`

tinmarino: :h expr5

For String concatenation ".." is preferred, since "." is ambiguous, it is also
used for |Dict| member access and floating point numbers.

tinmarino: Despite Vim recommendation, I prefer keeping same style. At some point in the future, we may replace all `.` by `..`

autoload/vimwiki/base.vim
```diff
@@ -325,7 +325,7 @@ function! vimwiki#base#system_open_link(url) abort
       else
         let url = shellescape(a:url, 1)
       endif
-      execute 'silent ! start ' . url
+      execute 'silent !start ' .. eval(url)
```
tinmarino, line 328: I think the use of eval here is a bug. why to you evaluate a string like "http://www.example.com" ?
Plus it makes Vimwiki vulnerable to Remote Code Execution.

Please remote it.

doc/vimwiki.txt
```diff
@@ -4038,6 +4039,12 @@ master is retained as a legacy mirror of the dev branch.
 This is somewhat experimental, and will probably be refined over time.
 
 
+2024.03.08~
+
+Fixed:~
+    * Issue #1343: Fix/Improvement: Following HTML link on Windows PowerShell
+      or cmd.exe was broken by change to |:!start| behavior.
+
```
tinmarino, line 4047: Good job for adding the release note.

tinmarino (changes requested):
Remove the `eval`function

tinmarino:
Hi @jfishe thank you for the PR. Please apply the comments:

* Remove the eval function
* Replace the double dot by single dot
* Fix Vint automatic test

tinmarino:
Relates to #1343 and #1392

autoload/vimwiki/base.vim
jfishe: shellescape(a:url, 1) always outputs a single quote string, which powershell and pwsh reject. To remove eval(), I replaced shellescape with escape(a:url,  '%|*#').

johannish:
I'm new to vimwiki, but I applied this change manually and I can confirm that it fixes #1392 for me (though Vim version is already `version 9.0.2003`).

Shell is cmd on Windows 11.

:VimwikiShowVersion
```
Version: 2024.01.24
Os: Windows
Vim: 900
Branch: dev
Revision: 69318e7
Date: 2024-03-16 14:23:28 -0300  
```

PR #1401, opened 2024-04-23
Fix tags generation when wiki extension has multiple dots

MartyLake:
When using extension containing multiple dots, for example `.wiki.txt`, generated tags would contains a faulty links that ends with piece of the extension, for example `.wiki`, then vimwiki would try to go to a file that ends with `.wiki.wiki.txt` instead.

This patch solves the issue by removing the exact suffix of the extension instead of using the root filename modifier.


MartyLake:
superceded by https://github.com/vimwiki/vimwiki/pull/1438/files

Issue #1445, opened 2024-10-03
Tests failing on dev branch


fancsali:
Looking into running the tests locally and fix the issue on my PRs, I discovered, that numerous tests are - in fact - failing on the `dev` branch!

Running the same tests as GitHub CI would, I get varying level of success:

```
Running version: vim_7.3.429
[...]
Success/Total: 677/685 (assertions: 742/745)
[...]
Running version: vim_8.1.0519
[...]
Success/Total: 677/685 (assertions: 787/790)
[...]
Running version: nvim_0.3.8
[...]
Success/Total: 679/685 (assertions: 793/796)
[...]
Running version: vim_7.4.1099
[...]
Success/Total: 119/119 (assertions: 174/174)
[...]
Running version: vim_8.0.0027
[...]
Success/Total: 321/322 (assertions: 494/494)
```

Full logs below:
[test.out.txt](https://github.com/user-attachments/files/17245924/test.out.txt)

fancsali:
Going back to **v2023.04.04_1** or **v2023.04.04** (which are the same, only the version number bump was missing from the code), we're _all green_! 

| Test/Version | Result (Success/Total) |
|--------|--------|
| vim_7.3.429 | 655/655 (assertions: 702/702) |
| vim_8.1.0519 | 655/655 (assertions: 747/747) |
| nvim_0.3.8 | 655/655 (assertions: 747/747) |
| vim_7.4.1099 | 97/97 (assertions: 124/124) |
| vim_7.4.1546 | 244/244 (assertions: 125/125)|
| vim_8.0.0027 | 314/314 (assertions: 495/495)| 

NOTE: The number of tests is somewhat less compared to the **v2024.01.22** and **v2024.01.24** versions.

fancsali:
In summary, the broken tests are:

- [ ] Link creation - `/testplugin/test/link_creation.vader:(5,6),40,47`
- [ ] Link renaming - /testplugin/test/link_renaming.vader:2,(6,9),14,16`
- [ ] Tag file format - `testplugin/test/tag.vader:16`

```
Running version: vim_7.3.429
=============================
  [...]
  Starting Vader: /testplugin/test/link_creation.vader
    (40/50) [ EXPECT] (X) Full Words are made as links, no extension addded . -> space
    (47/50) [ EXPECT] (X) Markdown links
  Success/Total: 48/50
  [...]
  Starting Vader: /testplugin/test/link_renaming.vader
[...]

PR #1446, opened 2024-10-14
test/tag.vader: update hardcoded version to 2024.01.24

brennen:
This should fix part of #1445.


brennen:
I'm going to go ahead and merge this on the theory that it doesn't make anything _worse_.

PR #1450, opened 2024-11-18
Fix: use relative path when opening wiki files

jpmor:
vimwiki constructs absolute paths internally (by concatenating `path + key + ext`
from `get_wikilocal`), so `edit_file` always receives an absolute path. This means
vimwiki buffers always display their full path in the statusline, even when working
inside the wiki directory — unlike standard Vim behavior when opening files directly.

This fix uses the `:.` filename modifier to reduce the path to be relative to the
current directory when possible, falling back to the absolute path otherwise.

The key ordering: `fnamemodify` is applied to the raw `a:filename` before
`fnameescape`, since `fnamemodify` expects a literal path. Applying it to the
already-escaped `fname` would silently fail for filenames with spaces or special
characters.

---

- [x] Make sure to update the documentation in `doc/vimwiki.txt` if applicable, including the Changelog and Contributors sections.

### DIFF
--- a/autoload/vimwiki/base.vim
+++ b/autoload/vimwiki/base.vim
@@ -1159,7 +1159,7 @@ function! vimwiki#base#edit_file(command, filename, anchor, ...) abort
   " getpos() directly after this command. Strange.
   if !(a:command =~# ':\?[ed].*' && vimwiki#path#is_equal(a:filename, expand('%:p')))
     try
-      execute a:command fname
+      execute a:command fnameescape(fnamemodify(a:filename, ':.'))
     catch /E37:/
       call vimwiki#u#warn('Can''t leave the current buffer, because it is modified. Hint: Take a look at'
             \ . ''':h g:vimwiki_autowriteall'' to see how to save automatically.')
--- a/doc/vimwiki.txt
+++ b/doc/vimwiki.txt
@@ -4019,6 +4019,7 @@ Contributors and their Github usernames in roughly chronological order:
     - @jiamingc
     - Alex Claman (@claman)
     - @qadzek
+    - @jpmor
 
 
 ==============================================================================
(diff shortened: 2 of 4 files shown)


brennen:
Could you expand on the actual difference in behavior here, and what the change fixes? Comments suggest this path is only executed if we're jumping anchors within a file (or at least if we're opening the current file) - does that seem right?

jpmor:
> Comments suggest this path is only executed if we're jumping anchors within a file (or at least if we're opening the current file) - does that seem right?
I think those comments describe this block being skipped for anchors within a file: https://github.com/vimwiki/vimwiki/blob/72792615e739d0eb54a9c8f7e0a46a6e2407c9e8/autoload/vimwiki/base.vim#L1160

> Could you expand on the actual difference in behavior here, and what the change fixes?
I prefer not to have my full path shown if I'm editing a file in my cwd. Ex: if I do `vim foo.wiki` it will open `foo.wiki`, but opening `foo.wiki` with vimwiki opens `/full/path/to/wikidir/foo.wiki`.

The `:.` filename-modifier behavior is just what I want... and edit_file seemed like a good central place to apply that.

jpmor:
Thanks for the patience on this. I've updated the commit with a fix to the ordering issue.

The original patch applied `fnamemodify` to `fname`, which is already the output of `fnameescape()`. Since `fnamemodify` expects a literal path, this could silently fail for filenames containing spaces or other special characters — the escaped backslashes would confuse the path comparison and prevent `:.` from reducing to a relative path.

The fix is to apply `fnamemodify` to the raw `a:filename` first, then escape the result:

```vim
execute a:command fnameescape(fnamemodify(a:filename, ':.'))
```

This matches the pattern already used on line 1147 where `fnamemodify` is correctly applied to `a:filename` directly.

To summarize the motivation: vimwiki constructs absolute paths by concatenating `path + key + ext` from `get_wikilocal`, so `edit_file` always receives an absolute path. Without this fix, vimwiki buffers always display with their full path even when working inside the wiki directory. The `:.` modifier degrades gracefully — if the file isn't under CWD, the path is left unchanged.

Happy to adjust anything. Is this ready to merge?

